PR #23, opened 2015-04-20
Fix to gui bug of Armor Smithy

LauraRozier:
Created the fix for the gui bug surrounding the Armor Smithy.
Made a value flip-flop at the render pool.
Also tidied up the fix as a request by Krom.
Here also the block diagram used during my bug tracing: http://www.gliffy.com/go/publish/image/7857823/L.png

### DIFF
--- a/src/KM_RenderPool.pas
+++ b/src/KM_RenderPool.pas
@@ -664,7 +664,7 @@ end;
 
 procedure TRenderPool.AddHouseSupply(aHouse: THouseType; Loc: TKMPoint; const R1, R2: array of Byte);
 var
-  Id,I,K: Integer;
+  Id,I,K,Ident1,Ident2: Integer;
   R: TRXData;
 
   procedure AddHouseSupplySprite(aId: Integer);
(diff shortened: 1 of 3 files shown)


LauraRozier:
Meanwhile I have found the source of the issue, yet I am not able to fix it in that location, it is in KaM it's  own houses.dat file, not in KaM Remake.
This fix is perfectly fine, until the houses.dat file is fixed, then it should to be put back to the original. :)

If there is no possibility to fix the .dat file then this is the ideal fix.
It requires less resources, processor wise and size wise, then an xor flip-flop like:
x = x xor y
y = x xor y
x = x xor y
For this to work properly I would need to apply more code and it would make little to no sense to apply the method if I can also do it with fewer lines of code and easier to understand logic for both human and computer.

I hope that I explained the edit well in this comment, with questions please do comment to this post, contact me or contact one of the project's programmers.

Kromster80:
Please edit the following code:

```
if aHouse = ht_ArmorSmithy then
begin
  if I = 1 then // If sprite 1 gets called, render house sprite 2 (Coal)
    Id := gRes.HouseDat[aHouse].SupplyIn[2, Min(R1[I - 1], 5)] + 1
  else if I = 2 then // If sprite 2 gets called, render house sprite 1 (Steel)
    Id := gRes.HouseDat[aHouse].SupplyIn[1, Min(R1[I - 1], 5)] + 1
  else // Back to normal processing for remaining two entries
    Id := gRes.HouseDat[aHouse].SupplyIn[I, Min(R1[I - 1], 5)] + 1;
end else // End of fix
  Id := gRes.HouseDat[aHouse].SupplyIn[I, Min(R1[I - 1], 5)] + 1;
```

into

```
Id := gRes.HouseDat[aHouse].SupplyIn[I, Min(R1[I - 1], 5)] + 1;

// Need to swap Coal and Steel for the ArmorSmithy
// For some reason KaM stores these wares in swapped order, here we fix it (1 <-> 2)
if (aHouse = ht_ArmorSmithy) and (I in [1,2]) then
  Id := gRes.HouseDat[aHouse].SupplyIn[3-I, Min(R1[I - 1], 5)] + 1;
```

It is more compact and clear that way. Same for SupplyOut

LauraRozier:
Will do, update under way.

PR #24, opened 2015-04-21
A fix for the crackling audio right after the loading screen.

LauraRozier:
Now fades in from 0 to volume setting.
Also set the fade time to 3000 to have a slightly more natural overflow of music.

### DIFF
--- a/src/KM_Music.pas
+++ b/src/KM_Music.pas
@@ -51,7 +51,6 @@ type
     constructor Create(aVolume: Single);
     destructor Destroy; override;
     procedure UpdateMusicVolume(Value: Single);
-    procedure StartMusicVolume; // Used to fix audio crackling after loading screen
     procedure PlayMenuTrack;
     procedure PlayNextTrack;
     procedure PlayPreviousTrack;
@@ -216,18 +215,6 @@ begin
 end;
 
 
-procedure TMusicLib.StartMusicVolume; // Used to fix audio crackling after loading screen
-begin
-  if not IsMusicInitialized then Exit; //Keep silent
-  {$IFDEF USELIBZPLAY}
-  ZPlayer.SetPlayerVolume(Round(0 * 100), Round(0 * 100)); //0=silent, 100=max
-  {$ENDIF}
-  {$IFDEF USEBASS}
-  BASS_ChannelSetAttribute(fBassStream, BASS_ATTRIB_VOL, 0); //0=silent, 1=max
-  {$ENDIF}
-end; // End if fix part
-
-
 procedure TMusicLib.ScanMusicTracks(aPath: UnicodeString);
 var
   SearchRec: TSearchRec;
@@ -277,15 +264,19 @@ end;
 
 
 procedure TMusicLib.PlayMenuTrack;
+var
+  remVolume: Single;
 begin
   if not IsMusicInitialized then Exit;
   if fMusicCount = 0 then Exit; //no music files found
   if fMusicIndex = 0 then Exit; //It's already playing
   fMusicIndex := 0;
+  // There was audio crackling after loading screen, here we fix it
+  remVolume := MusicGain; // Store volume setting to re-apply later on
+  MusicGain := 0; // Set audio level to 0
   PlayMusicFile(fMusicTracks[0]); // Must start playing before being able to adjust volume
-  // Used to fix audio crackling after loading screen
-  StartMusicVolume; // Set audio level to 0
-  UnfadeMusic(nil); // Slowly fade the music to set volume (3s)
+  MusicGain := remVolume; // Set audio level to stored volume setting
+  UnfadeMusic(nil); // Slowly fade-in to set volume (3s)
 end;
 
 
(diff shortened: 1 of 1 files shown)


LauraRozier:
Let me explain the case here.
I am using an external headphone amplifier as audio output to my headphones.
I have it connected to a USB2 port and a 240v AC power adapter to the amp to provide the power required.

Why I have this? I used to be an audio engineer and I mixed a ton of multi-tracks on my PC, using Reaper, which I love.
It is still a hobby of mine.

I never have any issues with audio, just when I start KaM Remake I do have a brief crackling in the music track, right after the loading screen.
I applied a little delay to the fade-in function, procedure TMusicLib.UnfadeMusic;, and it starts off at a volume level of 0 to then fade-in in 3000 ms, 3 seconds, to the volume set in the settings menu, which by default is 10, that's 0.5 for bass and 0.5*100=50 for libzplay.

With questions please do reply to this post, contact me or contact one of KaM Remake's programmers.

src/KM_Music.pas
```diff
@@ -264,12 +264,19 @@ procedure TMusicLib.ScanMusicTracks(aPath: UnicodeString);
 
 
 procedure TMusicLib.PlayMenuTrack;
+var
+  remVolume: Single;
```
Kromster80: Rename to prevVolume

src/KM_Music.pas
```diff
@@ -423,10 +430,12 @@ procedure TMusicLib.UnfadeMusic;
   ZPlayer.GetPosition(StartTime);
   EndTime.ms := StartTime.ms + FADE_TIME;
   ZPlayer.GetPlayerVolume(Left, Right); //Start fade from the current volume
+  Sleep(25); // Sleep for 25 ms, give the stream time to buffer a few bytes
```
Kromster80: Does this sleep(25) happens within main game loop, effectively suspending the whole game by 50ms?

LauraRozier:
If that is the case I will have to add a checking method.
I will let you know when I am home. :)
Also, the name change will be fixed then too.

src/KM_Music.pas
LauraRozier: Ah, so after checking it a bit, no just 25 ms, they are within the {$IFDEF USELIBZPLAY} ... {$ENDIF} and {$IFDEF USEBASS} ... {$ENDIF} tags.

I am going to add a Boolean called fStartState, this will be set to False at the start and then be set to True when PlayMenuTrack gets called, then the fade will check if the value is True and add the 25 ms delay, in the end it is set to False again so the 25 ms delay will not be used anymore unless it is set to True again.

src/KM_Music.pas
```diff
@@ -98,7 +100,7 @@ constructor TMusicLib.Create(aVolume: Single);
   end;
   {$ENDIF}
 
-  UpdateMusicVolume(aVolume);
+  UpdateMusicVolume(aVolume); // Must stay here, sets audio volume to fade to.
```
Kromster80: Why would it need to be removed from here?

src/KM_Music.pas
```diff
@@ -72,7 +73,7 @@ implementation
 uses
   KM_Log, KM_Utils;
 
-const FADE_TIME = 2000; //Time that a fade takes to occur in ms
+const FADE_TIME = 3000; //Time that a fade takes to occur in ms <- Set to 3000 for a nicer overflow of sound
```
Kromster80: Lets keep that as it was, 2sec sounds quite okay

src/KM_Music.pas
```diff
@@ -264,12 +266,20 @@ procedure TMusicLib.ScanMusicTracks(aPath: UnicodeString);
...
+var
+  prevVolume: Single;
 begin
   if not IsMusicInitialized then Exit;
   if fMusicCount = 0 then Exit; //no music files found
   if fMusicIndex = 0 then Exit; //It's already playing
   fMusicIndex := 0;
-  PlayMusicFile(fMusicTracks[0]);
+  // There was audio crackling after loading screen, here we fix it
+  prevVolume := MusicGain; // Store volume setting to re-apply later on
```
Kromster80: Comments on this and following lines are superfluous and can be removed - they repeat the code

src/KM_Music.pas
```diff
@@ -264,12 +266,20 @@ procedure TMusicLib.ScanMusicTracks(aPath: UnicodeString);
...
 procedure TMusicLib.PlayMenuTrack;
+var
+  prevVolume: Single;
 begin
   if not IsMusicInitialized then Exit;
   if fMusicCount = 0 then Exit; //no music files found
   if fMusicIndex = 0 then Exit; //It's already playing
   fMusicIndex := 0;
-  PlayMusicFile(fMusicTracks[0]);
+  // There was audio crackling after loading screen, here we fix it
```
Kromster80: "here we fix it by doing A-B-C"
This can be expanded with details

src/KM_Music.pas
```diff
@@ -423,10 +433,20 @@ procedure TMusicLib.UnfadeMusic;
...
+    Sleep(25); // Sleep for 25 ms, give the stream time to buffer a few bytes, only if USELIBZPLAY
+    fStartState := False; // Set fStartState to False so normal opperation returns
+  end;
   ZPlayer.SlideVolume(tfMillisecond, StartTime, Left, Right, tfMillisecond, EndTime, Round(MusicGain * 100), Round(MusicGain * 100));
   {$ENDIF}
   {$IFDEF USEBASS}
   BASS_ChannelPlay(fBassStream, False); //Music may have been paused due to fade out
+  if fStartState then
+  begin
+    Sleep(25); // Sleep for 25 ms, give the stream time to buffer a few bytes, only if USEBASS
```
Kromster80: "Sleep for 25 ms" in comments is superfluous

src/KM_Music.pas
```diff
@@ -423,10 +433,20 @@ procedure TMusicLib.UnfadeMusic;
...
+    fStartState := False; // Set fStartState to False so normal opperation returns
```
Kromster80: "Set fStartState to False" is superfluous

src/KM_Music.pas
LauraRozier: Old comment that I forgot to take out, wanted to set the volume to 0 there, but that would make no sense as it would overwrite the MusicGain value, that would be a no-go, plus it is useless to change something that has nothing to do with the issue.

PR #26, opened 2015-04-23
Possible fix for the player defeated message when someone intentionally leaves the game

LauraRozier:
The extra use is needed to get the quitting player's HandIndex.
Trying to find a better way so you won't get the notification that a player lost when the player kept playing after victory.

Not able to test it for myself as of right now, due to not having the KM_NetAuthSecure.pas file.

Please report back with your findings? :)

### DIFF
--- a/src/KM_GameApp.pas
+++ b/src/KM_GameApp.pas
@@ -6,7 +6,7 @@ uses
   {$IFDEF Unix} LCLType, {$ENDIF}
   {$IFDEF WDC} UITypes, {$ENDIF}
   Classes, Controls, Dialogs, ExtCtrls, KromUtils, Math, SysUtils, TypInfo,
-  KM_CommonTypes, KM_Defaults, KM_RenderControl,
+  KM_CommonTypes, KM_Defaults, KM_RenderControl, KM_HandsCollection,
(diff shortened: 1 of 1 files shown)


lewinjh:
That won't work. It only sets the player to defeated on the client who quits, other clients won't see anything different.

You don't need NetAuthSecure to test it. You can use the unsecure one that is in the repo by changing the flag in KaM_Remake.inc. There are instructions on the wiki:
https://github.com/Kromster80/kam_remake/wiki/ProjectCompilation

The issue is also more complex that just writing the code. Because if we implement that feature it becomes impossible to replace quitters. Currently if a player quits in MP you can return to the lobby and get someone else to replace them. But if we defeat them instantly then they can't be replaced. That's why I never implemented it, it needs further consideration before starting coding.

LauraRozier:
Got it, I will have a look at different points on the list instead.
Thanks for the reply. :)

Kromster80:
Yeah, some issues are much deeper than it looks on the first sight. Please contact us before picking a new one, to avoid such problems :)

LauraRozier:
Will do, time to put more time in reading the issues and code connected to the issues.

PR #27, opened 2015-04-25
Replay Rename

LauraRozier:
A fully functional and tested replay renaming function to add to KMR.
I have tested it with 3 replays: http://i.imgur.com/u4t3YXN.png

I made use of the Edit that only supports file name safe characters, to be sure that mistakes can't be made by accident.

I defined all the my edits in the commit notes, yet here is the complete list:
- Button added under the mini-map, gets enabled when an item is selected.
- A centered popup opens after clicking the rename button, the edit-field has the current name inside.
- The popup contains:
  - A fitting background image.
  - Two labels, one as a title and one in front of the edit-field.
  - One edit-field to change the name.
  - Two buttons, one to confirm and one to cancel.
- Added procedure RenameSave(...); to KM_Saves to allow the file name to be changed.
- Made KM_Saves more consistent in the variable sections).
- The list automatically updates and re-sorts the list.

Stuff to think about(+ improvement, - disadvantage):
- Updating only the edited replay instead of the full replay list.
  - \+ Would take less time to update the list.
  - \- Now it does re-sort the list.
  - \- Would require a new procedure just to get this right and take a lot of development time.

This is my very first full feature implementation for KMR, I hope I did well enough for a KMR-noob. ;)


lewinjh:
Hi, thanks for the pull request, your code looks good. I have a few suggestions though:
- The popup doesn't stay centered when resizing the screen. Set its Anchors to [] and it will stay centered (not anchored to anything).
- The background should be dimmed while the popup is active, as it is on the multiplayer screen for the Create Server popup.
- The rename button looks funny below the map, since it's separate from the other buttons but not related to the map. Maybe it can go in the blank space below the delete button? (and make it hide when the delete confirmation appears there).
- Text should be taken from the text library so it can be translated. Run TranslationManager.exe in the Remake folder, add the required text (in a similar format to existing text, for example Delete Replay), and commit the modified files (KM_TextIDs.inc and text.eng.libx).

Can you add those things to this pull request? Thanks!

Kromster80:
@lewinjh good ideas!
It would also make sense to put Rename and Delete on one line and make them both show pop-ups.

LauraRozier:
Will do, I now have it under delete and delete confirm is a popup as well.
Putting it next to delete might cause issues with other languages, I am not sure of the length of all the languages.

I will make the text in the text lib, I thought about it indeed and it would make a lot of sense.
Also the anchors will be fitted as requested. :)
And I will have a look at the dimming, shouldn't be a problem to fit that in as well.
Instead of hiding I will disable/enable the buttons according to what popup is opened/closed for a natural feeling. :)

Kromster80:
Sounds good

LauraRozier:
Little update, I have them completed now, just the text that supports translations, commit update will follow upon completion of that, for now preview images:
Delete Replay popup: http://i.imgur.com/XlYNrY0.png
Rename Replay popup: http://i.imgur.com/yxphqrw.png

LauraRozier:
One second, needs a tiny cleanup

LauraRozier:
There we go, ready for use. :)

lewinjh:
Looks good. One last request: The "confirm" button of the rename popup should be disabled when the file name entered already exists (i.e. renaming would overwrite an existing replay). This is the behaviour for the save menu in the game so you can check that.

LauraRozier:
Will look into it being added tomorrow, first a bit of sleep. :P

LauraRozier:
Done.

Kromster80:
Looks good to me. If there are any issues we can fix them in master.
Good work, thanks Thimo!

PR #40, opened 2015-08-12
Fixed adding Trees/Corn/Wine via dynamic script

vqw:
ScriptTryObjectSet fixed, checking for allowable objects as well (we do
not want falling trees etc)

### DIFF
--- a/src/terrain/KM_Terrain.pas
+++ b/src/terrain/KM_Terrain.pas
@@ -579,20 +579,21 @@ begin
 
   //Apply change
   //UpdatePassability and UpdateWalkConnect is used in SetField so that we only use it in trees and other objects
-  Land[Y, X].Obj := aObject;
   case aObject of
 
     55..58: // Wine in different stages
-      if CanAddField(X, Y, ft_Wine) then
+      if CanAddField(X, Y, ft_Wine) and (TileIsCoal(X, Y) <= 0) then // TileGoodForField does not check for coal deposit and puts a field there, we do not want this
       begin
+         Land[Y, X].Obj := aObject;
          SetField(KMPoint(X, Y), -1, ft_Wine);
          Result := True;
       end
       else
         Result := False;
     59..63: // Corn in different stages
-      if CanAddField(X, Y, ft_Corn) then
+      if CanAddField(X, Y, ft_Corn) and (TileIsCoal(X, Y) <= 0) then  // TileGoodForField does not check for coal deposit and puts a field there, we do not want this
       begin
+        Land[Y, X].Obj := aObject;
         SetField(KMPoint(X, Y), -1, ft_Corn);
         Result := True;
       end
@@ -601,6 +602,7 @@ begin
     88..124, // Trees - 125 is mushroom
     126..172:
     begin
+      Land[Y, X].Obj := aObject;
       if ObjectIsChopableTree(KMPoint(X,Y), caAge1) then Land[Y,X].TreeAge := 1;
       if ObjectIsChopableTree(KMPoint(X,Y), caAge2) then Land[Y,X].TreeAge := TREE_AGE_1;
       if ObjectIsChopableTree(KMPoint(X,Y), caAge3) then Land[Y,X].TreeAge := TREE_AGE_2;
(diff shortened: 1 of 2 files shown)


src/terrain/KM_Terrain.pas
```diff
@@ -548,6 +548,15 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
       end;
   end;
 
+  function AllowableObject : Boolean;
```
Kromster80, line 551: Please add a comment upfront, what is a "AllowableObject"

src/terrain/KM_Terrain.pas
```diff
@@ -555,20 +564,58 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
   DiagonalChanged := MapElem[Land[Y,X].Obj].DiagonalBlocked xor MapElem[aObject].DiagonalBlocked;
 
   //Apply change
   Land[Y, X].Obj := aObject;
-  UpdatePassability(KMRect(X, Y, X, Y)); //When using KMRect map bounds are checked by UpdatePassability
-  UpdateWalkConnect([wcWalk, wcRoad, wcWork], KMRectGrowTopLeft(KMRect(X, Y, X, Y)), DiagonalChanged);
-  Result := True;
+  case aObject of
+
+    55..58:
```
Kromster80: Please add comments on to each case statement, what are these magic numbers

vqw:
Comments added

src/terrain/KM_Terrain.pas
```diff
@@ -555,20 +564,59 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+      else
+        Result := False;
+    88..124, // Trees - 125 is mushroom
+    126..172:
+    begin
+      if ObjectIsChopableTree(KMPoint(X,Y), caAge1) then Land[Y,X].TreeAge := 1;
+      if ObjectIsChopableTree(KMPoint(X,Y), caAge2) then Land[Y,X].TreeAge := TREE_AGE_1;
+      if ObjectIsChopableTree(KMPoint(X,Y), caAge3) then Land[Y,X].TreeAge := TREE_AGE_2;
+      if ObjectIsChopableTree(KMPoint(X,Y), caAgeFull) then Land[Y,X].TreeAge := TREE_AGE_FULL;
+      UpdatePassability(KMRect(X, Y, X, Y)); //When using KMRect map bounds are checked by UpdatePassability
```
Kromster80: Why do we update passability and walk connect only for trees and other objects?

vqw:
UpdatePassability and UpdateWalkConnect is used in SetField so that we only use it in trees and other objects

src/terrain/KM_Terrain.pas
```diff
@@ -555,20 +564,60 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+      else
+        Result := False;
+    59..63: // Corn in different stages
+      if CanAddField(X, Y, ft_Corn) then
+      begin
+        SetField(KMPoint(X, Y), -1, ft_Corn);
+        Result := True;
+      end
+      else
+        Result := False;
```
Kromster80, line 601: When we return False, does this imply that the whole operation failed? And if it has failed, then maybe it should not have the "Land[Y, X].Obj := aObject;" set.

vqw:
Fixed this, along with placing Wine/Corn on coal deposits which was wrong.
TileGoodForField does not check this - did it manually in this corner case.

Kromster80:
When we return False from ScriptTryObjectSet, does this imply that the whole operation failed? And if it has failed, then maybe it should not have the "Land[Y, X].Obj := aObject;" set.

vqw:
"Land[Y, X].Obj := aObject;" moved inside each case statement (please check the src) and when cannot place field it returns False and means that whole operation failed.

Kromster80:
For the future, please create new pull requests separately.

PR #47, opened 2015-08-19
Fixed an issue causing Actions.AIGroupsFormationSet(..) to return bytecode error.

LauraRozier:
The issue was a bracket that shouldn't be there.


lewinjh:
Thanks!

PR #49, opened 2015-08-31
Rough first implementation

LauraRozier:
This is to implement timed messages for dynamic scripts.

Help is always welcome. ;P
It works for now and it's been tested from one to 100 messages and a check to determine whether it leaves the other messages be.
Test code used:

``` pascal
procedure OnMissionStart;
var
  i: Integer;
begin
  for i := 0 to 5 do
    Actions.ShowTimedMsg(0, 'This is a test message!' + IntToStr(i), 30);
  Actions.ShowMsg(0, 'This is a test message! to check if this stays');
  for i := 0 to 5 do
    Actions.ShowTimedMsgFormatted(0, 'This is a test message! %d', [i], 60);
end;

procedure OnTick;
begin
  if States.GameTime mod 120 = 0 then
    Actions.ShowTimedMsgFormatted(0, 'This is a test message! In-Game ticks = %d', [States.GameTime], 60);
end;
```

**ShowTimedMsg**
Purpose and use of this feature/function:
Shows a message to players for a certain amount of time in game ticks. If the player index is -1 the message will be shown to all players.
(Same as ShowMsg but with a timer.)

Parameters:
1 - player index 
2 - text (!AnsiString)
3 - time

**ShowTimedMsgFormatted**
Purpose and use of this feature/function:
Displays a message to a player with formatted arguments (same as Format function) and for a certain amount of time in game ticks. If the player index is -1 the message will be shown to all players.

Parameters:
1 - player index 
2 - text (!AnsiString) 
3 - Array of arguments
4 - time

### DIFF
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -986,6 +986,8 @@ end;
 
 procedure TKMGame.ShowTimedMessageLocal(aKind: TKMMessageKind; aText: UnicodeString; aLoc: TKMPoint; aTime: Integer);
 begin
+  // Add the message with the current tickcount plus the ticks stated in the script
+  // This way the timing is very accurate
   fGamePlayInterface.TimedMessageIssue(aKind, ParseTextMarkup(aText), aLoc, fGameTickCount + aTime);
 end;
 
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -241,7 +241,6 @@ begin
   gProjectiles := TKMProjectiles.Create;
 
   fGameTickCount := 0; //Restart counter
-  SetLength(fGamePlayInterface.fTimedMsg, 0);
 end;
 
 
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -1534,8 +1534,11 @@ begin
                       //Don't autosave if the game was put on hold during this tick
                       if fGameTickCount mod 600 = 0 then
                       begin
-                        if IsMultiplayer and fNetworking.IsHost then
-                            fGameInputProcess.CmdGame(gic_GameAutoSave, UTCNow) //Timestamp must be synchronised
+                        if IsMultiplayer then
+                        begin
+                          if fNetworking.IsHost then
+                            fGameInputProcess.CmdGame(gic_GameAutoSave, UTCNow); //Timestamp must be synchronised
+                        end
                         else
                           if gGameApp.GameSettings.Autosave then
                             fGameInputProcess.CmdGame(gic_GameAutoSave, UTCNow);
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -113,6 +113,7 @@ type
     procedure ShowMessage(aKind: TKMMessageKind; aTextID: Integer; aLoc: TKMPoint; aHandIndex: TKMHandIndex);
     procedure ShowMessageLocal(aKind: TKMMessageKind; aText: UnicodeString; aLoc: TKMPoint);
     procedure ShowMessageLocalFormatted(aKind: TKMMessageKind; aText: UnicodeString; aLoc: TKMPoint; aParams: array of const);
+    procedure ShowTimedMessageLocal(aKind: TKMMessageKind; aText: UnicodeString; aLoc: TKMPoint; aTime: Integer);
     procedure OverlayUpdate;
     procedure OverlaySet(const aText: UnicodeString; aPlayer: Shortint);
     procedure OverlaySetFormatted(const aText: UnicodeString; aParams: array of const; aPlayer: Shortint);
@@ -240,6 +241,7 @@ begin
   gProjectiles := TKMProjectiles.Create;
 
   fGameTickCount := 0; //Restart counter
+  SetLength(fGamePlayInterface.fTimedMsg, 0);
 end;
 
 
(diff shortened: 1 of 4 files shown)


src/KM_Game.pas
```diff
@@ -1513,16 +1520,21 @@ procedure TKMGame.UpdateGame(Sender: TObject);
...
                             fGameInputProcess.CmdGame(gic_GameAutoSave, UTCNow) //Timestamp must be synchronised
-                        end
                         else
                           if gGameApp.GameSettings.Autosave then
                             fGameInputProcess.CmdGame(gic_GameAutoSave, UTCNow);
                       end;
 
+
+                      for J := High(fGamePlayInterface.fTimedMsg) downto 0 do
+                        if Length(fGamePlayInterface.fTimedMsg) >= 0 then
```
Kromster80: What does this line do?

Kromster80:
Please outline the purpose and use of this feature/function

LauraRozier:
Will do.
Had to hurry to the academy, hence the "quick" work.
I'll have some time left there, will add proper short comments to explain the added/changed code.

thibmo:
Purpose and use of this feature/function:
Shows a message to players for a certain amount of time in game ticks. If the player index is -1 the message will be shown to all players.
(Same as ShowMsg but with a timer.)

Parameters:
1 - player index 
2 - text (!AnsiString)
3 - time

It has been requested and accepted (DynamicScripting toDo-list line 94)
For now it is only the bald "ShowMsg" kind, versions supporting formatting will follow asap (approx this weekend in the same branch). :)

thibmo:
**ShowTimedMsg**
Purpose and use of this feature/function:
Shows a message to players for a certain amount of time in game ticks. If the player index is -1 the message will be shown to all players.
(Same as ShowMsg but with a timer.)

Parameters:
1 - player index 
2 - text (!AnsiString)
3 - time

**ShowTimedMsgFormatted**
Purpose and use of this feature/function:
Displays a message to a player with formatted arguments (same as Format function) and for a certain amount of time in game ticks. If the player index is -1 the message will be shown to all players.

Parameters:
1 - player index 
2 - text (!AnsiString) 
3 - Array of arguments
4 - time

It has been requested and accepted (DynamicScripting toDo-list line 94)
For now it is only the bald "ShowMsg" kind, version supporting formatting will follow asap (approx this weekend in the same branch). :)

LauraRozier:
I think this would do. :)

lewinjh:
That was added to the accepted list ages ago, before we added the overlay text. The idea was to display message text overlaid on the screen for a certain time, such as "3rd wave attack is coming!" in tower defence. It wasn't intended to use the messages that stack up on the left side of the GUI (like ShowMsg). The game should not delete these stacked message automatically, that's inconsistent and confusing for the user since the user has control over when to delete them.

So the intention was actually something closer to the overlay text, but with stack of messages with timeouts. I'm not sure how it fits into the system now that we have the overlay, since that space/idea is already taken. Some scripts implemented their own timeout system to display messages on the overlay for a certain timeout (checkout my "Center Castle Looting" script for an example). The feature can be implemented purely on the script side and gives more flexibility since you can put other text above and below the messages. It could still be added to the API to make it easier for script authors because it is a common feature to want.

We really appreciate your contributions to the project, but unfortunately just because something is on the accepted list doesn't mean anyone has agreed how it will work. We just agreed that it is a good idea. Also things get added to accepted and then forgotten while other features supersede it (as the overlay text has kind of done in this case).

Perhaps you can change this to use the overlay? We could have a command
Actions.OverlayTextAppendTimedMessages(PlayerIndex)
which adds the timed messages to the overlay one per line. Then the script author can still add their own text above and below on the overlay, and this feature simply offers a convenience of not having to write the timeout system in their script.

We wouldn't need the GoTo variants.

Usage would be something like:

```
    procedure OnTick;
    begin
        Actions.OverlayTextSet(-1, '');
        Actions.OverlayTextAppendTimedMessages(-1);
    end;
```

(the script might add text such as player scores before or after the messages, but this is a simple example)
And to display a messages:

```
    AddTimedMsg(-1, '3rd wave attack is coming!', 100); //Will appear on the overlay for 10 seconds
```

What do you think?

src/KM_Game.pas
```diff
@@ -1513,16 +1534,20 @@ procedure TKMGame.UpdateGame(Sender: TObject);
                       //Don't autosave if the game was put on hold during this tick
                       if fGameTickCount mod 600 = 0 then
                       begin
-                        if IsMultiplayer then
-                        begin
-                          if fNetworking.IsHost then
+                        if IsMultiplayer and fNetworking.IsHost then
```
lewinjh: Why did you change the logic for saving games? The change you made means that ALL clients will send the gic_GameAutoSave command (if they have autosave enabled in settings) and thus in a game with 8 players the autosave will be created 8 times on each client! :-)

Only the host is supposed to send this command since they coordinate autosaves to be synchronised.

Kromster80:
I have to agree with Lewin. There is a danger in taking the todo items into implementation without discussing them with us. For another matter (adding to Lewins arguments) todo item text is just 40-60 symbols and can not possibly fit all the details and constraints. It is more of a reminder, than a detailed spec.

Let's outline the usage scenario first, how will it need look from the script side, e.g. can scripter add 50 timed messages interveaved with usual messages?

LauraRozier:
Concerning main topic:
@lewinjh After reading it like that it seems logical to use the overlay or create something close to the overlay(maybe a small box of 200x500 WxH in px?) 
@Kromster80 I should have first consulted you two for the global idea first, that is a fact, I just didn't want to keep you both from your work. :)
Right now indeed the scripter can create 50 messages that can be mixed in any way and the code should remove the right messages.
I did see that messages with identical texts do get mixed up (as I decided to check on exact text)

Concerning the edit that accidentally made it's way in this branch:
That was a small edit I made as I saw there is nothing else in the statement, I thought the "and" would result in the same. After re-reading I do understand the issue being caused by this. :P

PR #56, opened 2015-10-17
PascalScript update.

LauraRozier:
Small update that adds availability for Android and some stability improvements.
Checked with several maps and there seem to be no issues.


Kromster80:
Where do these changes come from?

LauraRozier:
They are from the original update.

Kromster80:
After discussion, we decided to not merge it. PascalScript works fine, no need to fix it. We also don't need any Android support either.

PR #62, opened 2015-12-12
Update - GUI changes and code changes

LauraRozier:
-  Removed the top-menu
- Added 2 input textboxes
-  Added 2 output textboxes
-  Reworked the layout
-  Now outputs 3 files rather then 1 combined file
-  No file-save dialog
-  File-open dialog stays for convenience, only opens when no inputs are set when parse is pressed
-  Removed the help-form
-  Supports relative paths
-  Saves settings to INI file

New looks:  
New looks


Kromster80:
Very nice, thank you! :-)

PR #63, opened 2015-12-13
Update/Cleanup - Removed unused procedure and reworked crash workaround

LauraRozier:
-  Tool crashed (Access Violation) on-load and fixed that by coupling OnChange via code, now it's done in the DFM
-  Set a boolean to solve the crash issue
-  Removed an unused procedure
-  Tweaked the layout to make it look cleaner
-  Few code cleanups


Kromster80:
Very well :-)

I have made some changes in the tool, make sure to get them!

PR #69, opened 2015-12-20
Small parser update

LauraRozier:
States had some cases where it used TIntegerArray, gets fixed here along with two forgotten types


Kromster80:
Merged, nice )

Issue #74, opened 2016-03-06
Dynamic script to change house/unit owner


Kromster80:
> I am surprised there is no such script? Imo it would be a very basic script... I know I can silently kill/destroy a unit/group/house and replace it with one from another player... But that's pretty ugly.

The Dark Lord

LauraRozier:
I was thinking about the following:

``` Pascal
Actions.HouseOwnerChange(aHouseID: Integer; aNewOwnerID: Shortint);
Actions.UnitOwnerChange(aUnitID: Integer; aNewOwnerID: Shortint);
Actions.GroupOwnerChange(aGroupID: Integer; aNewOwnerID: Shortint);
```

Then codes alike:

``` Pascal
procedure TKMScriptActions.HouseOwnerChange(aHouseID: Integer; aNewOwnerID: Shortint);
var
  aHouse: TKMHouse;
begin
  try
    if aHouseID > 0 then
    begin
      aHouse := fIDCache.GetHouse(aHouseID);
      if aHouse <> nil then
        aHouse.OwnerUpdate(aNewOwnerID);
    end else
      LogParamWarning('Actions.HouseOwnerChange', [aHouseID, aNewOwnerID]);
  except
    gScriptEvents.ExceptionOutsideScript := True; //Don't blame script for this exception
    raise;
  end;
end;

procedure TKMScriptActions.UnitOwnerChange(aUnitID: Integer; aNewOwnerID: Shortint);
var
  aUnit: TKMUnit;
begin
  try
    if aUnitID > 0 then
    begin
      aUnit := fIDCache.GetUnit(aUnitID);
      if aUnit <> nil then
        aUnit.SetOwner(aNewOwnerID);
    end else
      LogParamWarning('Actions.UnitOwnerChange', [aUnitID, aNewOwnerID]);
  except
    gScriptEvents.ExceptionOutsideScript := True; //Don't blame script for this exception
    raise;
  end;
end;

procedure TKMScriptActions.GroupOwnerChange(aGroupID: Integer; aNewOwnerID: Shortint);
var
  aGroup: TKMUnitGroup;
begin
  try
    if aGroupID > 0 then
    begin
      aGroup := fIDCache.GetUnit(aGroupID);
      if aGroup <> nil then
        aGroup.SetOwner(aNewOwnerID);
    end else
      LogParamWarning('Actions.UnitOwnerChange', [aGroupID, aNewOwnerID]);
  except
[...]

PR #75, opened 2016-03-07
[WIP] Owner changing via scripts

LauraRozier:
This PR brings what was requested in issue " Dynamic script to change house/unit owner #74 ".
Tested with the included map.

Might still require attention as hands also contain refs to unit/group/house pointers?


Kromster80:
This needs much more work and checking. I would not recommend to implement all 3 at once.

lewinjh:
There are lots of tricky cases to consider here which could cause unexpected situations, bugs or crashes:
- Units eating inside inn, inn owner is changed
- Unit working in house or out in fields, house owner is change but unit continues to live/work there even though it belongs to a different team
- Unit working in house or out in fields, unit owner changes, unit continues to live/work in house owned by different team
- Serf is carrying goods to a house, house owner is changed
- Houses register their supply/demand in delivery queue. House owner changes. House supply/demand is still registered in delivery queue for the wrong team. Deliveries to/from house still happen by serfs for old team not new team. Crashes will probably occur when the house tries to update the supply/demand.
- Serf is performing delivery, serf owner changes, serf's task's delivery ID is now invalid in the new team's delivery queue
- Recruit living in barracks, recruit owner is changed
- Labourer waiting for stone to build road, labourer owner is changed
- Two groups of soldiers are fighting, owners are changed to same team, units are now fighting units from their own team
- Food has been ordered for soldiers, soldiers team is changed
- etc. etc.

You could even have enemy units living inside your houses if they were changed to an enemy team.

The reason we didn't implement this previously and told people to "destroy and recreate" is because there's so many cases like this to solve. It's much neater to destroy the old unit/house and create a new one. All these loose ends are tied up through the destruction of the old unit/house.

I'm not sure it's worth trying to solve all of these cases since it will make the code more complicated/messy/hard to maintain (would we always have to assume everywhere that ownership can change at a moment's notice? Hard to do). Maybe you could implement a clean "destroy and recreate" solution as script actions?

LauraRozier:
That might be a better way indeed.
Silent destroy followed by a create at the same exact coordinates/state/etc.
One thing to think about then is, how to handle goods inside houses?
Just discard them or re-add them?

Kromster80:
Recreating state is often impossible, since it depends on many external variables, as Lewin extensively pointed out. For houses, e.g. I suppose only basic properties could be passed by (wares, damage).

If you are inclined to make this PR work, then I'd remove anything but the houses for start. Also note that scripting class should do as little as possible. Method likely should go into TKMHands.HouseOwhenChange(aOldOwner, aNewOwner: TPlayerIndex);

LauraRozier:
The most recent push only contains the HouseOwnerChange.
It stores and re-uses the following values:

``` Pascal
houseDmg: Word;
housePos: TKMPoint;
houseType: THouseType;
houseState: THouseState;
```

Now I just need to find a way to store the wares.
Though for Barracks and warehouses this might become complicated again.

LauraRozier:
Maybe this would be something that could be used?
Sample code image

Kromster80:
The most recent push contains all the resulting changes made in this PR (including units and groups functions you have added).

What does HouseState store?

LauraRozier:
My bad, this should be THouseBuildState, not THouseState.
Will be updated in the next push.
I should have also stated that the other ones are commented out, so technically still there.

To be honest I am not sure if one would even want to allow the house to be transferred when the TKMHouse.BuildingState is not 'hbs_Done'.

Right now the code doesn't allow the player to.

Kromster80:
I'm closing this PR for reasons well outlined by Lewin.

PR #76, opened 2016-04-16
Fixed Lazarus project file (fixing build paths)

ghost:
I've had couple of issues compiling the project in Lazarus 1.6, I had to fix build paths in project options (Common directory instead of src\common).


Kromster80:
Thanks for the fix! :-)

PR #94, opened 2016-11-01
Ability to change train queue priorities in school

reyandme:
Added hotkeys on Train button:
Shift+Left mouse btn - train unit immidiately. If other unit is
training, cancel it, move to pos 1 in queue, train current unit.
Ctrl+Left mouse btn - add unit in queue on 1st position (it will be
trained after current training unit)
In the queue bar same hotkeys were added: Shift+LMB - immidiate train.
Ctrl+LMB put 1st in queue.


Kromster80:
Please clean up the PullRequest. It should include only the feature it says to include (School train order). Other changes are superfluous.

PR #97, opened 2016-11-01
Delete field plans by holding left mouse button

reyandme:


src/KM_Defaults.pas
```diff
@@ -260,6 +260,14 @@ interface
     cmMarkers //CenterScreen, Defence, FOW markers
     );
 
+type
+  TKMCursorFieldMode = (
```
Kromster80, line 264: Comments would help here, explaining what this enum type represents

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -2822,6 +2850,19 @@ procedure TKMGamePlayInterface.MouseDown(Button: TMouseButton; Shift: TShiftStat
 // 2. Perform SelectingTroopDirection if it is active
 // 3. Display various cursors depending on whats below (might be called often)
 procedure TKMGamePlayInterface.MouseMove(Shift: TShiftState; X,Y: Integer);
+  procedure HandleFieldLMBDrag(P: TKMPoint; aFieldType: TFieldType);
+  begin
+    if not KMSamePoint(LastDragPoint, P) then
+      if (gMySpectator.Hand.CanAddFakeFieldPlan(P, aFieldType)) and (gGameCursor.Tag1 = Byte(cfmPlan)) then
```
Kromster80, line 2856: This is largely my mistake in the past. `Byte(something)` should be really `Ord(something)` throughout the code.

src/hands/KM_Hand.pas
```diff
@@ -472,37 +474,46 @@ procedure TKMHand.AddField(aLoc: TKMPoint; aFieldType: TFieldType);
 end;
 
 
+function TKMHand.LocHasNoAllyPlans(aLoc: TKMPoint): Boolean;
+var I: Byte;
```
Kromster80, line 478: Change Byte to Integer please, as discussed in other PR.

reyandme:
Find differ in behaviour with current version - now if u start placing road in the fog of war, then drag out of it - u will able to place road (only where FOW is reaveled).
after this feature pull request you are not able to do that, so you have to start from tiles, where FOW is revealed.
I think need to fix this.

reyandme:
Also I think we need to rename gic_BuildAddFieldPlan to gic_BuildToggleFieldPlan, because we actually toggle field plan state, but not add. Name gic_BuildAddFieldPlan confused me at the first moment.

src/gui/KM_InterfaceGamePlay.pas
reyandme: looks same here. But I can change to Ord here, as well as in other lines in this file and in RenderPool

reyandme:
To avoid multi commits as in another PR I will squash commits in local rep first, then publish polished version.

Kromster80:
Merged through TortoiseGit (applied as patch)

PR #104, opened 2016-11-02
Update Buglist.txt

reyandme:
Change unit training order in School on Ctrl+click, Shift+click is not a Bug, but Improvement :)


Kromster80:
Right, thanks! )

PR #105, opened 2016-11-02
Allow to block wares delivery into armor workshop

reyandme:
so that player could focus each on certain task (shields or leather armors)

### DIFF
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -329,8 +329,8 @@ type
   end;
 
 
-  {FlatButton}
-  TKMButtonFlat = class(TKMControl)
+  {Common Flat Button}
+  TKMButtonFlatCommon = class(TKMControl)
   private
   public
     RX: TRXType;
@@ -349,6 +349,15 @@ type
 
     procedure MouseUp(X,Y: Integer; Shift: TShiftState; Button: TMouseButton); override;
 
+    procedure Paint; override;
+
+  end;
+
+
+  {FlatButton}
+  TKMButtonFlat = class(TKMButtonFlatCommon)
+  private
+  public
     procedure Paint; override;
   end;
 
@@ -497,13 +506,11 @@ type
 
 
   {Row with resource name and icons}
-  TKMWaresRow = class(TKMControl)
+  TKMWaresRow = class(TKMButtonFlatCommon)
   public
-    RX: TRXType;
-    TexID: Word;
-    Caption: UnicodeString;
     WareCount: Byte;
-    constructor Create(aParent: TKMPanel; aLeft, aTop, aWidth: Integer);
+    constructor Create(aParent: TKMPanel; aLeft, aTop, aWidth: Integer); overload;
+    constructor Create(aParent: TKMPanel; aLeft, aTop, aWidth: Integer; aClickable: Boolean); overload;
     procedure Paint; override;
   end;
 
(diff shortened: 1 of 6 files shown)


src/gui/pages_game/KM_GUIGameHouse.pas
```diff
@@ -891,6 +992,27 @@ procedure TKMGUIGameHouse.House_SchoolUnitChange(Sender: TObject; Shift: TShiftS
 end;
 
 
+{Toggle ware delivery for separate resources (wood, leather) in Armor workshop}
+procedure TKMGUIGameHouse.House_ArmorWSDeliveryToggle(Sender: TObject);
+var I: Byte;
```
Kromster80: I should be Integer in almost all iterations

src/gui/pages_game/KM_GUIGameHouse.pas
```diff
@@ -891,6 +992,27 @@ procedure TKMGUIGameHouse.House_SchoolUnitChange(Sender: TObject; Shift: TShiftS
 end;
 
 
+{Toggle ware delivery for separate resources (wood, leather) in Armor workshop}
+procedure TKMGUIGameHouse.House_ArmorWSDeliveryToggle(Sender: TObject);
+var I: Byte;
+    WareType: TWareType;
+    ArmorWS: TKMHouseArmorWorkshop;
+begin
+  WareType := wt_None;
```
Kromster80: It would be much simpler to assign buttons `Tag := I;` in creation and here just map from `const WORKSHOP_WARE: array [1..2] of TWareType = (wtWood, wtLeather); ..... WareType := WORKSHOP_WARE[Sender.Tag]`

src/KM_Controls.pas
```diff
@@ -350,6 +350,15 @@   TKMButtonFlat = class(TKMControl)
     procedure MouseUp(X,Y: Integer; Shift: TShiftState; Button: TMouseButton); override;
 
     procedure Paint; override;
+
+  end;
+
+
+  {FlatButton}
+  TKMButtonFlat = class(TKMButtonFlatCommon)
```
Kromster80, line 358: Why is this needed? Every control has already OnClicked event that we could use.

Kromster80 (changes requested):

src/KM_Controls.pas
reyandme: TKMWaresRow inherits same behaviour as TKMButtonFlat. Sound, highliting etc. So i extract common parent for both of them. The only difference - painting. 
Do you think its better just to copy paste ?

src/gui/pages_game/KM_GUIGameHouse.pas
reyandme: +

reyandme:
@Kromster80 here it is, how it looks like:
https://puu.sh/svnXl/00dca688b3.png
Looks absolutely same as for Storehouse or Barracks

reyandme:
Месяц ожидания. 2й пошел.
Month of waiting. 2nd just have started.

Kromster80:
@lewinjh I don't quite like the function duplication between "ware not allowed"-red-triangle and "house does not allow wares in"-doors-closed-button.
For Barracks/Store it is kind of justified. Workshops are in between - they take multiple wares unliek other houses, yet still use "ware-lines" UI.
What's your opinion on this one?

lewinjh:
@Kromster80 The armour workshop is the only house where this feature makes sense since other houses that take multiple wares require both to make anything. So it's really a special case for barracks, store and armour workshop. It might be adding unnecessary complexity given that you can restrict wood delivery to all armour workshops in ware distribution (so if you plan to not make shields you can set wood to armour workshops to zero).

Kromster80:
@lewinjh so what would be the decision? I'm inclining towards "reject"

lewinjh:
@Kromster80 Yeah me too, it's a special case for a single house and basically just an optimisation to save a few wood if you don't plan to make shields in one of your armoury workshops. I'm ok with rejecting it if you are.

PR #109, opened 2016-11-04
After separate single unit of group, select it

reyandme:
We did split to use him, so selecting is very usefull


Kromster80:
Sounds like a good idea, thanks!

PR #120, opened 2016-11-08
Food exchange rates fix

reyandme:
Fixes only rates from food to smth else (except food), one way only.
To prevent players from trading starting food in storehouse to other good to get fast weapons. Normally food should be used to feed, not to get fast weapons. With these rates we make it very unprofitable.

Also small graphic fix - remove superfluous colon in the rates info labels


Kromster80:
How exactly this solves the problem mentioned above?
If player gets 10 axes instead of 20, he is still in a good position to rush.

PR #122, opened 2016-11-09
Fixed objects placement via scripts

andreus791:
+ Mapmakers can use object 255 to clear specified tiles.
+ No need to force placing corn/wine fields in manner like here ( SetField(KMPoint(X, Y), -1, ft_Wine); ). Mapmaker can use grapes/corn as decoration, then fields aren't needed. There's also Actions.GiveField scripts for this.

### DIFF
--- a/src/terrain/KM_Terrain.pas
+++ b/src/terrain/KM_Terrain.pas
@@ -580,6 +580,22 @@ begin
   //Apply change
   //UpdatePassability and UpdateWalkConnect are called in SetField so that we only use it in trees and other objects
   case aObject of
+    55..59: // Corn and grapes: Set FieldAge if there's field to let them grow
+              begin
+                Land[Y, X].Obj := aObject;
+                Result := True;
+                if TileIsWineField(KMPoint(X, Y)) then
+                case aObject of
+                  55: Land[Y,X].FieldAge := WINE_AGE_1;
+                  56: Land[Y,X].FieldAge := WINE_AGE_2;
+                  57: Land[Y,X].FieldAge := WINE_AGE_FULL - 1;
+                end;
+                if TileIsCornField(KMPoint(X, Y)) then
+                case aObject of
+                  58: Land[Y,X].FieldAge := CORN_AGE_2;
+                  59: Land[Y,X].FieldAge := CORN_AGE_FULL - 1;
+                end;
+              end;
     88..124,
     126..172: // Trees - 125 is mushroom
               begin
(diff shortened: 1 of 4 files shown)


lewinjh:
We already have a button in the map editor for deleting objects, why do we need to have object 255 in the list?

Allowing placing field/wine objects without fields sounds ok to me.

andreus791:
It is needed for dynamic scripts.

I tried to allow all objects in MapEd to see how it works, but the list ended at object 251. Object 255 was still hidden. So I added it to "scripting" list.

lewinjh:
I see, yes object 255 and 61 should be allowed from scripts :)

lewinjh:
However, your change to allow placing corn/wine on fields won't work, since at the start of ScriptTryObjectSet it blocks placing objects on fields.

andreus791:
Yeah, maybe it should be reworked to smth like "if tile is occupied by worker then exit". Occupied means worker already has a task and goes to that tile, or works on that tile. For corn, wine, trees. Used trees for example :D
https://youtu.be/bCiVMfB25gw

andreus791:
"Occupied means worker already has a task and goes to that tile, or works on that tile."
Like this. Occupied loc is marked as brown tile.

https://www.youtube.com/watch?v=EG4zZTpx0Ss

lewinjh:
Is it really necessary to add a boolean to the terrain tile record just for this minor scripting feature? It also adds a lot of complexity keeping that boolean up to date which will be hard to maintain. I don't like this solution.

Does it matter if the script author changes an object on the tile where a unit will be working? As long as the game doesn't crash I don't think it matters if for example the woodcutter plants there anyway and replaces the object. If the script is magically making objects appear that is the script author's choice. As far as I am aware the game handles this situation anyway, the woodcutter will plant somewhere else if an object is placed where he plans to work?

For fields you need to be careful that the script can't mess up the state of the field. Previously we blocked placing objects on fields since that's a simple way to prevent it being messed up. I would be happy for you to allow placing the corn objects on fields (and wine on winefields) to cause the field to grow to that stage (but not other objects obviously).

Maybe make a separate pull request for your "allow object 255" commit. I am very happy to accept that one since it's actually a bug (we intended setting object 255 to clear the current object, that's what it says it does in the scripting documentation).

Kromster80:
I agree with Lewin.

andreus791:
Woodcutter/Farmer will just find another place to work, like if there were 2-3 units going to cut the same tree or harvest grapes from the same winefield.

Let it be as it is now ;)

PR #125, opened 2016-11-12
Map name label

andreus791:
image

### DIFF
--- a/src/gui/KM_InterfaceDefaults.pas
+++ b/src/gui/KM_InterfaceDefaults.pas
@@ -56,10 +56,6 @@ const
   OPT_SLIDER_MIN = 0;
   OPT_SLIDER_MAX = 20;
   MAX_SAVENAME_LENGTH = 50;
-  //Limit names length to fit interface width
-  MAX_MAPNAME_LENGTH = 22;
-  MAX_TRACKNAME_LENGTH = 18;
-
 
 implementation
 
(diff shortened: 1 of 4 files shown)


lewinjh:
It makes the interface look too crowded in my opinion.

Kromster80:
As suggested on Trello:
- song name could go down between buttons (as it was in KaM iirc). 
- also take into account that whole menu must reasonably fit into smallest game window (1024*720 iirc).

andreus791:
image

andreus791:
default
default

Limited map name length (.InCastle 1.3 (Bonus)_ Barb_Balanced_77FE8E32 --> .InCastle 1.3 (Bonus)_), dynamic interface to fit autowrapped track names.

Kromster80:
Let's make the texts uniform:

Game time // yellow
hh:mm:ss // white

Map title // yellow
my_map_name // white

Music player // yellow
01_Busy // white

andreus791:
Done, looks better now

image

src/gui/KM_InterfaceDefaults.pas
```diff
@@ -56,6 +56,9 @@   TKMUserInterfaceCommon = class
   OPT_SLIDER_MIN = 0;
   OPT_SLIDER_MAX = 20;
   MAX_SAVENAME_LENGTH = 50;
+  //Limit names length to fit interface width
+  MAX_MAPNAME_LENGTH = 22;
+  MAX_TRACKNAME_LENGTH = 18;
```
Kromster80: Let's move these 2 consts to a module where they are used. No need to make them "globally" visible.
After that PR can be merged )

Kromster80 (changes requested):

Kromster80:
Thanks!

PR #129, opened 2016-11-14
Updated GiveField/GiveWinefield scripts

andreus791:
Mapmaker can set field growth stage. Finally, Corn/Wine placed to fields will grow.
image


lewinjh:
This changes the scripting API, which means any scripts written for r6720 that use these API commands will no longer work and need to be updated. It would be better to not change existing commands, and instead add new commands for this functionality (for example Actions.SetFieldAged).

Breaking API compatibility is something we should avoid doing unless there is a very good reason.

andreus791:
Hmm, adding extra script commands without breaking API compatibility sounds like a good solution.

lewinjh:
Thanks :)

PR #130, opened 2016-11-15
Lazarus compilation fixes

reyandme:
But app do not start on x64 win7 machine with error 0xc000007b

### DIFF
--- a/src/common/pascalscript/Source/uPSRuntime.pas
+++ b/src/common/pascalscript/Source/uPSRuntime.pas
@@ -4227,7 +4227,7 @@ begin
     end;
     GetMem(darr, Longint(NewLength * elSize) + SizeOf(TDynArrayRecHeader));
     {$IFDEF CPUX64}
-    darr^.header._Padding:=0;
+    //0xdarr^.header._Padding:=0; //We ignore x64 machines
     {$ENDIF CPUX64}
     darr^.header.refCnt:=1;
     {$IFDEF FPC}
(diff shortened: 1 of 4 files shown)


src/common/pascalscript/Source/uPSRuntime.pas
```diff
@@ -4227,7 +4227,7 @@ procedure PSDynArraySetLength(var arr: Pointer; aType: TPSTypeRec; NewLength: Lo
     end;
     GetMem(darr, Longint(NewLength * elSize) + SizeOf(TDynArrayRecHeader));
     {$IFDEF CPUX64}
-    darr^.header._Padding:=0;
+    //0xdarr^.header._Padding:=0; //We ignore x64 machines
```
Kromster80, line 4230: Why is it prefixed //0x ?

Kromster80 (changes requested):

src/common/pascalscript/Source/uPSRuntime.pas
reyandme: have no idea... Lazarus editor. Anyway it was compilation error there - smth like "no property _Padding for object".

Kromster80: Why did it enter there if it is surrounded with CPUX64 ? 

Please read this messages last paragraph: http://www.knightsandmerchants.net/forum/viewtopic.php?p=38197&sid=b62bf461852f14aa3f1056e593eedb88#p38197

It looks like you have x64 Lazarus with x64 target. KMR should be built with x32 for x32

Kromster80 (changes requested):

Issue #20, opened 2015-03-31
Bug: Actions.UnitKill and invalid groups


Kistrel:
Using Actions.UnitKill on a newly equipped Warrior before he joined a group creates an invalid group that exists and is alive, but has 0 members and is hungry.

Kromster80:
Please describe the algorithm of reproducing this bug, in steps.

Kistrel:
Create a map in the editor, with a Barracks stocked with recruits and axes.
Attach the following script, then start a game on the map and recruit a Militia.

```
procedure OnTick;
var
  aUnits: array of Integer;
  i: Integer;
begin
  if States.StatUnitTypeCount(0, 14) > 0 then begin
    aUnits := States.PlayerGetAllUnits(0);
    for i := 0 to Length(aUnits)-1 do if States.UnitType(aUnits[i]) = 14 then
      Actions.UnitKill(aUnits[i], True);
  end;
end;
```

PR #132, opened 2016-11-16
"Unit is hungry" message could be shown for invalid/dead groups

andreus791:
As described by Kistrel, see issue #20.

### DIFF
--- a/src/units/KM_UnitGroups.pas
+++ b/src/units/KM_UnitGroups.pas
@@ -1433,12 +1433,14 @@ begin
 
   SomeoneHungry := False;
   for I := 0 to Count - 1 do
-  begin
-    SomeoneHungry := SomeoneHungry
-                     or ((Members[I].Condition < UNIT_MIN_CONDITION)
-                     and not Members[I].RequestedFood);
-    if SomeoneHungry then Break;
-  end;
+    if (Members[I] <> nil) 
+    and not Members[I].IsDeadOrDying
+    begin
+      SomeoneHungry := SomeoneHungry
+                       or ((Members[I].Condition < UNIT_MIN_CONDITION)
+                       and not Members[I].RequestedFood);
+      if SomeoneHungry then Break;
+    end;
 
   if SomeoneHungry then
   begin


lewinjh:
Thanks! Does this resolve the underlying problem behind Kistrel's issue though? Does the TKMGroup eventually get removed or is there another bug where the unit dies but you end up with an empty group that sticks around?

andreus791:
Seems group is alive because it has at least one member. This member is dying and still can be accessed. So game checks dying members (they probably have condition and another attributes <= 0) and sends message. The same thing happens sometimes when unit dies in game because of hunger (if hunger message check and death happen at the same tick, player gets message). Checked code - any valid group should be removed when dies in UpdateState.

Then checked with actions.log
previous script, plus:

procedure onwarriorequipped(u,g:integer);
begin
  Actions.Log('Groups successfully created. Group ID: ' + IntToStr(g) + '; Members: ' + IntToStr(States.GroupMemberCount(g)) + '; Member IDs: ' + IntToStr(u));
end;

procedure OnUnitDied(id,ak:integer);
begin
  Actions.Log('Group dies. Group ID: ' + IntToStr(states.unitsgroup(id)) + '; Member IDs: ' + IntToStr(id));
end;

procedure onplanroadplaced(p,x,y:integer);
var i:integer; gg:array of integer;
begin
actions.planremove(p,x,y);
gg := states.playergetallgroups(0);
actions.log('Current state: ' + inttostr(length(gg)) + ' groups found');
if length(gg) > 0 then
for i := 0 to length(gg) - 1 do
actions.log('Groups: ' + inttostr(gg[i]));
end;

2016.11.17 10:54:03.645   Current state: 0 groups found
2016.11.17 10:54:06.284   Group dies. Group ID: -1; Member IDs: 12058160
2016.11.17 10:54:07.371   Groups successfully created. Group ID: 4046371; Members: 1; Member IDs: 12058160
2016.11.17 10:54:09.774   Current state: 1 groups found
2016.11.17 10:54:09.774   Groups: 4046371

Group is saved somewhere in array and can be accessed, f.e in states.groupdead script.
According to log, group dies before it was created.

reyandme:
Pls fix this:
KM_UnitGroups.pas(1438,5) Fatal: Syntax error, "THEN" expected but "BEGIN" found

andreus791:
fixed in PR #134

Issue #15, opened 2015-03-31
Suggest last entered room password when creating a new room


Kromster80:
Maybe game could keep the last entered password and show it as a default option when creating a room.

PR #133, opened 2016-11-16
Allow to save and use last entered password

andreus791:
#15
Maybe similar option for lobby description could be useful.

### DIFF
--- a/KM_TextIDs.inc
+++ b/KM_TextIDs.inc
@@ -690,6 +690,7 @@ TX_LOBBY_DOWNLOAD_OVERWRITE = 738; //Overwrite existing files
 TX_LOBBY_RESET_BANS = 768; //Reset all bans
 TX_LOBBY_VIEW_README = 825; //View Readme PDF
 TX_LOBBY_PDF_ERROR = 826; //Failed to launch PDF viewer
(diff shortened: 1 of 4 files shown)


Kromster80:
Storing plane passwords sounds like a bad idea. I don't recall other games offering such function.

Kromster80:
@lewinjh what do you say?

lewinjh:
Well these aren't private account passwords, just a codeword shared amongst friends to let you enter the room, so security isn't such a big issue. I wouldn't expect people to reuse their Gmail password here since they have to tell it to their friends ;)

As for the feature, I don't really have an objection. I guess if you're often creating lobbies for the same group of friends it would be annoying typing in the password every time? (I assume you implemented it because it's a feature you'd like to see as a player)

andreus791:
As a player I often create rooms with the same password. And yes, I'd like to see this feature

Kromster80:
Okay then, I'm in :-)

Kromster80:
Maybe it is better to rename the label / variables to "Remember password"
Also immediately erase password when player unchecks the checkbox

Kromster80:
@andreus791 please see if you can do this part too: "Also immediately erase password when player unchecks the checkbox"

PR #136, opened 2016-11-19
Option to automatically pause the replay when peacetime ends

andreus791:


Kromster80:
Will this work well if a game is on a speedup?

See this line within gGame.Update: "for I := 1 to fGameSpeedMultiplier do"
that icreases current tick.
I'm also not happy with a separate checkibox in UI, maybe it could be
hidden (be just a flag in INI, disabled by default?)

reyandme:
conslusion after discussion on skype:
add option in menu (next to "autosave"). In Replay mode also show in mini options page on F4

Kromster80:
And check/fix the pause on speedup

src/KM_Game.pas
```diff
@@ -1561,7 +1562,23 @@ procedure TKMGame.UpdateGame(Sender: TObject);
                     end;
 
                     //Break the for loop (if we are using speed up)
-                    if DoGameHold then break;
+                    if DoGameHold then
+                      Break;
+
+                    PeaceTicksRemaining := Max(0, Int64((fGameOptions.Peacetime * 600)) - fGameTickCount);
```
Kromster80: Why do we need to cast explicitly to Int64?

src/KM_Game.pas
```diff
@@ -1457,6 +1457,7 @@ procedure TKMGame.Load(const aPathName: UnicodeString);
 procedure TKMGame.UpdateGame(Sender: TObject);
 var
   I: Integer;
+  PeaceTicksRemaining: Cardinal;
```
Kromster80: Better rename to `PeaceTimeLeft` - so we have keywords PeaceTime

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -983,7 +984,8 @@ procedure TKMGamePlayInterface.Create_Replay;
     Dropbox_ReplayFOW.OnChange := ReplayClick;
     Checkbox_ReplayFOW := TKMCheckBox.Create(Panel_ReplayFOW, 0, 25, 220, 20, gResTexts[TX_REPLAY_SHOW_FOG], fnt_Metal);
     Checkbox_ReplayFOW.OnClick := ReplayClick;
-end;
+
```
Kromster80: Unneeded line

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -1723,13 +1725,15 @@ procedure TKMGamePlayInterface.AlliesUpdateMapping;
 end;
 
 
+procedure TKMGamePlayInterface.SetButtons(aPaused: Boolean);
```
Kromster80, line 1727: Why do we need this function to become accessible from anywhere else?

src/gui/pages_game/KM_GUIGameMenuSettings.pas
```diff
@@ -45,40 +46,46 @@ constructor TKMGameMenuSettings.Create(aParent: TKMPanel);
   Panel_Settings := TKMPanel.Create(aParent, TB_PAD, 44, TB_WIDTH, 332);
     CheckBox_Settings_Autosave := TKMCheckBox.Create(Panel_Settings,PAD,15,WID,20,gResTexts[TX_MENU_OPTIONS_AUTOSAVE],fnt_Metal);
     CheckBox_Settings_Autosave.OnClick := Menu_Settings_Change;
-    TrackBar_Settings_Brightness := TKMTrackBar.Create(Panel_Settings,PAD,40,WID,0,20);
+    CheckBox_Settings_ReplayAutoPause := TKMCheckBox.Create(Panel_Settings,PAD,40,WID,20,'Replay autopause',fnt_Metal); //TODO translate
+    CheckBox_Settings_ReplayAutoPause.Hint := 'Automatically pause replay when peacetime ends'; //TODO translate
```
Kromster80: Please stick to Delphi format, including : ,  when adding todos like so "//todo:"

Kromster80 (changes requested):

reyandme:
Now 'Replay autopause option' is visible (but not enabled) during any game in game settings menu. I think we do need to show it (even in disabled state) unless its MP replay game mode.

Also in KM_GUIGameMenuSettings.pas we need to extend height of Panel_Settings with extra 25px.

Here instead of height 332 we need 357:
`Panel_Settings := TKMPanel.Create(aParent, TB_PAD, 44, TB_WIDTH, 332);`

I can fix this in new PR

PR #137, opened 2016-11-20
BugFix: voting return to lobby goes wrong if 2 or more last votes comes simultaneously

reyandme:
I couldn't fix this bug without proper trace logging on sending/receiving packets.


lewinjh:
Nice work! I've seen the bug occur while I was playing, didn't know this was the reason.

PR #138, opened 2016-11-24
Bugfix minimap borders inaccuracies

reyandme:
3 small visual bugs with minimap fixed
also improved coord mapping on minimap

check trello card https://trello.com/c/mkEO45l8


lewinjh:
Nice fixes! We never really checked edge cases like these with the minimap, we just got it working "good enough". Thanks :)

PR #139, opened 2016-11-24
Hotkeys bugfixes and improvements

reyandme:
1. 	Do not allow to set Ctrl or Shift as key bindings
2. 	Let use Tab for bindings
3. 	Add ability to cancel changes of keys bindings
4. If Object assigned to another key - clear previous assignation
5. New hotkeys for Plan road/field/wine/erase plans (for game and for MapEd)
6. Added 'Clear key' button in options menu
7. New Hotkey to switch selection between same type buildings/units/groups
8. Added support of extra mouse buttons (forward/backward) as hotkeys
9. Added hotkeys to switch between players while watching replay
10. Added hotkey to set game speed in replay as it was during the game (before/after PT)
11. KM_ResKeys refactoring


reyandme:
@lewinjh What do you think about these feactures ?
One thing I am not sure here - mb set no default keys for new extra 10 dynamic hotkeys (currently I set default keys as Numpad 1-9,0 same as for current 10 synamic hotkeys, but on Numpad.

src/gui/pages_menu/KM_GUIMenuOptions.pas
```diff
@@ -59,19 +61,21 @@   TKMMenuOptions = class { (TKMGUIPage) }
...
         ColumnBox_OptionsKeys: TKMColumnBox;
+        Button_OptionsKeysClear: TKMButton;
         Button_OptionsKeysReset: TKMButton;
         Button_OptionsKeysOK: TKMButton;
+        Button_OptionsKeysCancel: TKMButton;
       Button_OptionsBack: TKMButton;
   public
     constructor Create(aParent: TKMPanel; aOnPageChange: TGUIEventText);
-
+    destructor Destroy;
```
Kromster80: destructors must have "override;" keyword after them

src/hands/KM_Hand.pas
```diff
@@ -79,7 +79,7 @@   TKMHand = class(TKMHandCommon)
     InCinematic: Boolean;
 
     //Used for syncing hotkeys in multiplayer saves only. UI keeps local value to avoid GIP delays
-    SelectionHotkeys: array[0..9] of Integer;
+    SelectionHotkeys: array[0..19] of Integer;
```
Kromster80: Where does this magic number comes from? Const would be better

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -2361,7 +2370,7 @@ procedure TKMGamePlayInterface.Selection_Select(aId: Word);
   if gMySpectator.Hand.InCinematic then
     Exit;
 
-  if not InRange(aId, 0, 9) then Exit;
+  if not InRange(aId, 0, length(fSelection) - 1) then Exit;
```
Kromster80: "length(fSelection) - 1" usually can be replaced with "High(fSelection)"

Kromster80 (changes requested):

src/hands/KM_Hand.pas
reyandme: definitely

Kromster80:
This PR is out of date and contains a lot of different changes. Please see if you can actualize it and split into several manageable chunks.

reyandme:
I've actualized this PR.

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -2783,15 +2841,53 @@ procedure TKMGamePlayInterface.KeyUp(Key: Word; Shift: TShiftState);
...
+    if (SpecPlayerIndex <> -1) and (Dropbox_ReplayFOW.Count >= SpecPlayerIndex) then
+    begin
+      Dropbox_ReplayFOW.ItemIndex := SpecPlayerIndex - 1;
+      ReplayClick(Dropbox_ReplayFOW);
+      ReplayKeyPressed := True;
+    end;
+
+    if ReplayKeyPressed then
+      // 'Spec/replay' type button was clicked, so we have to Exit here, because it we could invoke 'Game' type button as well
+      Exit;
```
lewinjh: Can be simplified by moving Exit above

src/KM_Game.pas
```diff
@@ -75,6 +75,7 @@   TKMGame = class
     DoGameHold: Boolean; //Request to run GameHold after UpdateState has finished
     DoGameHoldState: TGameResultMsg; //The type of GameHold we want to occur due to DoGameHold
     SkipReplayEndCheck: Boolean;
+    IsDefaultGameSpeedInReplayMP: Boolean;  // While watching MP replay we can set to default game speed as it was in MP game
```
lewinjh: I don't think this is necessary. If you've sped up the replay why should it suddenly return to the default speed at peacetime? The viewer can do that themselves.

lewinjh (changes requested):
Lets get this monster PR merged! Changes look good overall. Fixes these few things and I'll merge it.

One thought I had (for the future, this PR is big enough!): It would be nice to keep the keybindings compatible between updates (especially with frequent beta testing). This means we can't reorder items in KM_KeyIDs.inc, otherwise the bindings will be assigned to the wrong action after an update. Maybe give each area an offset (e.g. +1000 for faCommon, +2000 for faGame, etc.) so you can add more keys to the end of each section without rearranging. But obviously the fFuncs array shouldn't have 1000s of elements just because of these offsets, so you'd need to find a way to keep it compact (maybe by separating key IDs from fFunc indexes).

reyandme:
@lewinjh Ok, I removed SC_REAL_GAME_SPEED hotkey as well as IsDefaultGameSpeedInReplayMP.

fFunc refactoring is better to do in separate PR, this one is tooooo big^)

lewinjh (approved):

PR #140, opened 2016-11-25
bugfix: new host is not rdy, after initial host was disconnected

reyandme:
if host disconnects in game, then after returning to lobby new host is not rdy

### DIFF
--- a/src/net/KM_Networking.pas
+++ b/src/net/KM_Networking.pas
@@ -635,6 +635,12 @@ begin
 
   fSelectGameKind := ngk_Save;
 
+  fMyIndex := fNetPlayers.NiknameToLocal(fMyNikname); // host's index can change when players are removed
+  fHostIndex := fMyIndex;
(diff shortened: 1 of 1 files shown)


lewinjh:
Should you call SendPlayerListAndRefreshPlayersSetup again after setting IsReady and HasMapOrSave? Otherwise other clients won't know about the change so they'll still show the host as not ready? I might be wrong, it's been many years since I last worked on this code ;)

reyandme:
Actually in this case we call SendPlayerListAndRefreshPlayersSetup 2nd time in TKMGameApp.StopGameReturnToLobby (have no idea why need to do it twice btw ;), but who knows in other cases... So I've done some refactoring here

lewinjh:
That looks like a neater solution. I suggest retesting it just in case there's some issue with it we haven't thought of.

KM_Networking.pas is a bit of a mess and in need of refactoring, mostly my fault since I wrote a lot of it xD
Apologies if it hurts your brain trying to understand the logic :)

PR #141, opened 2016-11-26
MapEd new feature: ability to change owner for objects

reyandme:
for units/houses/unit groups

Also 	MapEd have 2 black colors for teams. Replace one black with purple and little bit reorder colors for better visibility in MapEd


Kromster80:
Please don't put two separate changes into one Pull Request.

I guess we could change default KaM colors set now, as backwards color compatibility is quite unimportant IMO.

Changing objects owners in MapEd needs a more thorough reviewing.

Kromster80:
This looks really confusing in MapEd. A paintbucket icon would be much better and would not need special "TKMFlatButton2TriangesShape" class. Also it would be good to change mouse cursor into said paintbucket when it is in the "team_changing" mode.

image

reyandme:
Ok, I asked Malin to help us with new icon and cursor for this feature

reyandme:
Some refactoring: 
1. renamed aUpdateOwnerList into aMoveToNewOwner
2. removed TKMFlatButton2TriangesShape, use TKMButtonFlat instead

Kromster80:
Okay, waiting for new icon/cursor then

reyandme:
Cursor:
Cursor

Icon:
Icon

reyandme:
We can show cursor when press button and start 'PaintBucket' mode, or we can show it over the house/unit which we want to 'paint' (while in 'PaintBucket' mode). Not sure what is better. I personally like 2nd option more, because 

Also we could use mask (ink flowing from a bucket) for Cursor to show what color are we using at the moment.

In Photoshop, when you choose paintbucket to use, cursor is still default (arrow), until you left tool palette area or menu. May be we can do same here - cursor become paintbucket only on the map, not on the menu area (left panel)

@Kromster80 @lewinjh what do you think ?

reyandme:
Finally I have finished this feature. What we have now:

1. New button with new icon, new cursor for "change owner" feature. New "paintbucket" cursor appears only on the map. RMB - cancel "change owner" mode
2. able to change owner of house/unit/group/road/corn/wine
3. on mouse over highlight object to change owner: for units/groups and houses. Change flag color to a new one and apply new flag color mask to object (enlighten 2 times, with restrictions on brightness 0.3-0.9 to avoid absolutely black or white masks). For road/field/wine - show cyan quad.
4. Holding mouse and move - color everything under the mouse

Here is a small gif how does it work:
[link to gif](https://puu.sh/tXxQG/0b4889aac7.gif)

There is small viual bug, when house, group or unit is highlighted then it can be rendered over other map objects (like in gif - knights over tower). But may be its ok, because its better to highlight what are we going to color. And currently if we want to delete unit or terrain object we have same issue.

@Kromster80 said it looks ok on skype
@lewinjh what do you say ?

lewinjh:
Looks ok to me. Highlighted objects appearing over others is fine IMO.

Kromster80:
Looks okay to me too.

PR #142, opened 2016-11-26
Change default player colors.

reyandme:
no need to stay with original KaM colors


reyandme:
May be completely switch to our colors in MP_TEAM_COLORS ?
Because our, let say, red in MP_TEAM_COLORS, differs from KaM original hence DefaultTeamColors.
So we can put here our Red, Cyan, Green, Magenta, Yellow, Grey, Black, Orange, Blue, Light green, White and one other color (purple, for example)

lewinjh:
I agree it makes no sense keeping these default colours in the editor.

The reason for using the original colours was because some KaM campaign missions (TSK / TPR) don't specify the colours, but rely on the defaults. So with this change the black AI in TSK mission 20 might now be orange for example.

The best solution is probably checking that all of the TSK/TPR .dat files in KaM Remake do not rely on these defaults (and if they do set the colours and commit the fixed .dat files). Alternatively the .dat parser could use different default colours to the map editor.

Kromster80:
I'd suggest to include the "color checking" code into the Batcher util (one
that we use to resave and other mass operations on all of the maps)

lewinjh:
I forgot about batcher! That's the right place for it.

Kromster80:
@reyandme please see the Batcher code and add a button there to mass-check all the maps (campaign, sp, mp, tutorials). All colors must be set explicitly. If they are not set - apply default TSK colors and resave.

reyandme:
@Kromster80 ok

Kromster80:
@reyandme Did you have a chance to update the Batcher and all stock maps to have colors listed explicitly?

reyandme:
@Kromster80 I never checked Batcher code and what is going on here so its a low priority thing for me now.

Kromster80:
Please check the Batcher and see if you can make a function in there to check/update all the KMR maps. Then this PR will be good to accept.

reyandme:
I took a look at the KMR color system finally. 

We have 3 commands to set player color in *.dat file: 
!SET_MAP_COLOR
!SET_RGB_COLOR
!SET_NEW_REMAP

The last one (!SET_NEW_REMAP) is disused. Btw it still exists in 1 MP map, about 10 SP Maps, and 3 Campaigns: TPS,TSK and Neryn.
Suggestion 1: delete all these !SET_NEW_REMAP from maps (manually by Notepad++,f.e., or with Batcher another new function)

!SET_MAP_COLOR sets player color by 0-255 index from palette. There are several palettes, by I could only find usage of DefaultPalette, which is actually comes from file '/data/gfx/pal_0.bbm'.
I do not know is there any real use of other *.bbm files, and also of *.lbm files (not sure what is the difference, btw)
Suggestion 2: do not load unused palettes. Am I missing something ?
Question 3: it's not hard to get palette's contents, but do we have any kind of editor for them ? 

!SET_RGB_COLOR command sets player color to specified RGB value (which is written in decimal form, not hex form). If there are both commands (SET_MAP_COLOR and SET_RGB_COLOR), then the last command overrides previous one. MapEditor add both of them, when you save the game, so SET_MAP_COLOR is always unused. I do not know, why we need to add both, not just use SET_RGB_COLOR? Compatibility with original KaM ? We should parse them properly, ok, but there is no need to save them for new map.

Suggestion 4: Do not add SET_MAP_COLOR when save game in MapEditor.

And finally about using Default colors. 

As I understand, there is always either SET_MAP_COLOR or SET_RGB_COLOR or both. If there is SET_RGB_COLOR - then everything is fine, we get the color by its rgb code.
If there is *only* SET_MAP_COLOR, then it will be taken from palette, which is always the same, then the color will be fine as well.

@Kromster80 or @lewinjh please explain, what exaclty should this new function do in batcher for all SP/MP/Campaign maps?

Should we get rid of DefaultPalette at all ?

What I was talking about in the first comment is to get rid of KM_Defaults.DefaultTeamColors which are used as default colors in MapEd, and they have 2 black colors.

lewinjh:
Suggestion 1: This was used for minimap colour in KaM. Removing it is fine by me, although since it's ignored it's not really necessary to delete it.

Suggestion 2: The palette used to be needed for loading .rx files, that might not be needed in the main KaM Remake application now. Before you remove anything check that our tools/utils aren't using them either, for example RXXPacker needs to load other palettes IIRC (and I think it shares the same code).

Question 3: Palettes are not supposed to be changed. They're an old way of compressing sprites to only use 1 byte per pixel (both on disk and on GPU) by only having 256 colours available, chosen from a palette to suit the game's needs. We now convert everything to RGB during loading or in RXX packer. We only use palettes for !SET_MAP_COLOR to be able to load KaM mission files correctly. It uses the same palette as the .rx sprites were compressed with.

Suggestion 4: Agreed, this is a leftover from before we implemented !SET_RGB_COLOR. It can be removed now.

@reyandme The problem is if the .dat file does not use !SET_MAP_COLOR or !SET_RGB_COLOR for a player. In KaM it uses a colour from the default list of colours, so to maintain compatibility this is also what we do. If you change the default list of colours, this behaviour will be broken and so original KaM campaigns could load with the wrong player colours in cases where they did not specify but relied on the default colours. So if you want to change these default colours you need to make the batcher check all existing missions for teams that do not have a colour specified and add it. Or just change the default list that the map editor uses when you create a blank map, so that old KaM missions still load correctly.

reyandme:
Suggestion 1: not a big deal, but we parse this unused line :)
Suggestion 2: I believe they could be used by RXXPacker, so I do not want to remove their usage, but as you said, they are not needed for main game (not need waste time for loading unused resources, even if they not so much "time expensive")
Question 3: I do not want to edit them, but it would be nice to at least easily read them. Anyway its not a problem at all, just asking
Suggestion 4: Ok

Ok, I got it, I think it should be not difficult to add this func to batcher

reyandme:
New 4 buttons added to Batcher:
1. Check all players in all maps have at least one of commands SET_MAP_COLOR or SET_RGB_COLOR
2. For every player, who do have any of these commands add SET_RGB_COLOR command
3. Delete command SET_NEW_REMAP from all maps (unused) (suggestion 1 impl)
4. Some maps have both SET_MAP_COLOR and SET_RGB_COLOR, SET_MAP_COLOR is always coming first, means SET_RGB_COLOR is always overriding color previously set by SET_MAP_COLOR. 4th button finds all such maps and deletes SET_MAP_COLOR command. (suggestion 4 impl)

So lots of maps are going to change after this. So its quit difficult to track changes in the maps.
I asked @Kromster80 and he suggested to unxor all the maps, because there is no need in this kind of "protection" anymore - anything can be edited via MapEditor and its even much simplier. @lewinjh what do think, can we do that ?

lewinjh:
@reyandme Yes unxoring the maps is fine by me. Maybe first unxor them, then apply these changes so that the changes are visible in github?

Does the map editor write unxored files now? If not it should.

Utils/Batcher/Unit1.dfm
```diff
@@ -180,4 +179,67 @@ object Form1: TForm1
...
+    OnClick = btnCheckColorClick
+  end
+  object btnSetDefColor: TButton
+    Left = 7
+    Top = 310
+    Width = 258
+    Height = 25
+    Hint = 
+      'Check if there is color specified in *.dat file for every player' +
+      ' by either SET_MAP_COLOR or SET_RGB_COLOR command'
```
lewinjh: Hint is copied from above

Utils/Batcher/Unit1.pas
```diff
@@ -42,20 +66,47 @@   TForm1 = class(TForm)
...
+    $FF27A700, //Green
+    $FFFF67FF, //Magenta
+    $FF07FFFF, //Yellow
+    $FF577B7B, //Grey
+    $FF2383FB, //Orange
+    $FFFF0707, //Blue
+    $FF0BE73F, //Light green
+    $FF720468, //Purple
+    $FFFFFFFF, //White
+    $FF000000  //Black
```
lewinjh: Looks like you copied your changed version by mistake. The original KaM default colours had black repeated.

lewinjh (changes requested):

lewinjh (approved):

lewinjh:
Thanks. Code changes all look good. Do you want to sort out the XORing and commit the .dat changes in this PR too?

reyandme:
> Does the map editor write unxored files now? If not it should.

I fixed this.

Last commit - I unXORed all maps (289 in total). In next PR we can apply *.dat files changes by new buttons I added here.

reyandme:
I decided to revert all maps unxor here, to be able to see other changes. For unxored maps there is another PR #278

PR #143, opened 2016-11-26
BugFix: it's possible to type in spec channel in lobby chat, when not in specs anymore

reyandme:
When first come as spec, choose spec channel, left specs (select any loc or random), spec channel still active and u can type in it, but can not see anything.


lewinjh:
Nice catch

PR #144, opened 2016-11-27
Warriors separate game statistics in MP

reyandme:


lewinjh:
I don't like having ut_ArmyPower, it's not a unit. Also who decides these power rates for each soldier? Seems a bit subjective to me.

Could you post a screenshot of the results page with the new army chart selected?

Kromster80:
ut_ArmyPower is certainly a "no go"

reyandme:
Krom it was originally your idea actually. So I implemented it. Nobody else like it ...

Kromster80:
I don't recall ever suggesting to introduce a new "ut_ArmyPower" type, can
you check please?

reyandme:
It was ur idea about adding this feature. So Lewin asked about what is power rates.
What about wt_Food and other ware types, which are used especially in stats ?
We have "special" unit types like ut_none and ut_any - which are not units also. But they looks usefull

reyandme:
@lewinjh @Kromster80 this is how it looks
https://puu.sh/sygT4/acaeebcd6a.png

Kromster80:
wt_Food is used in army feeding, it is a generic food type, because
warriors do not know what exact food they will get when the serf takes it
from Store.
Also it makes sense = TWareType-Food. TUnitType-ArmyPower is not logical )

Kromster80:
Just in case, the rest (idea and chart) look good to me :-)

reyandme:
what if we make smth like this:

TUnitTypeAll = (ut_None, ut_Any,
    ut_Serf,          ut_Woodcutter,    ut_Miner,         ut_AnimalBreeder,
... all unit types here...
    ut_Crab,         ut_Waterflower,  ut_Waterleaf,    ut_Duck,
uta_ArmyPower);

  TUnitType = ut_None..ut_Duck;

Actually then its better to separate all "real" unit types and "special" ones, not only uta_ArmyPower. But may be its too much testing, then...

Kromster80:
See, ut_ArmyPower is not an Unit Type at all, not even an aggregate one. So
it's better to rework chart display and add there dedicated list of
entitites and links.

reyandme:
I dont get it. Neither ut_Any nor ut_None.

Kromster80:
Any, None, Warriors, Animals, etc, are unit type agregates (true, not the
best way to store them, but it saves a lot of hassle).
UnitPower is completely different entity from unit type.

lewinjh:
The screenshot looks great! Nice work.

Maybe remove the heading "Number of warriors:" and replace:
 - "In total" with "Total Equipped"
 - "At the moment" with "Instantaneous"

reyandme:
>>Maybe remove the heading "Number of warriors:" and replace:
"In total" with "Total Equipped"
"At the moment" with "Instantaneous"

That is what I was struggling of - have no idea how to name it correctly ^)

What about label - I think it still can be usefull in case we have no good or too long translation. We are limited in length here - margin on the right is the same as on the left. On screenshot "number of warriors" is almost the max possible length here.

So we can left label on the same place, but set it empty for english translation.

What about ArmyPower - I will try to refactor code.
Also we need to set unit power rates somehow. I set them to some numbers, on my own.

lewinjh:
Merged :)

I'm not opposed to adding the army power to the chart but I agree it shouldn't be a unit type. Feel free to make a pull request for that.

lewinjh:
Also, @reyandme could you please make a screenshot with more interesting data in the lines? I'd like to post it to our social media as a teaser for fans to let them know that the game is still being developed and improved by a different team.

reyandme:
@lewinjh check this out: https://puu.sh/sWS50/39f0733133.gif

PR #145, opened 2016-12-03
Mute option

reyandme:


Kromster80:
Please write explanation when making a PR
What is this option for?
How it is different from setting sound volume to 0 ?

reyandme:
> What is this option for?

Just to mute both sound and music at one time, withuot clearing sound/music volume balance. 
Also it could be usefull with a hotkey on it. So if you want to mute game for a while you can press mute hotkey, then press again and sound/music volumes will be at the same volume levels, as they were before, no need for reajustments.
> How it is different from setting sound volume to 0 ? 

Technically it is setting music/sound volumes to 0. Unfortunally I do not know is it possible to pause music, if yes, then its probably better to pause music instead of silenting it.

reyandme:
PR updated to actual state

lewinjh:
I'm not convinced this is a necessary feature, as Krom said you can just set the volume to zero. Is it useful to have a hotkey to mute sound? I can't think of other games that have that.

reyandme:
Well, may be we can add this hotkey, without having extra mute option ?
So it will just set musicoff and sound volume to 0.

lewinjh:
Would anyone use that hotkey though? Do other games have a mute hot key?

reyandme:
I will :) Do not about others. Not sure about other games, but we can do something better, then others :)

Ok, may be this is not so important anyway. We can mute globally with some external software or just turn off speakers.

PR #146, opened 2016-12-06
Market ware blocked for trade until house which can produce it is build

reyandme:
mb need to change the texture of locked Item from question mark to lock or smth else.

### DIFF
--- a/src/KM_Defaults.pas
+++ b/src/KM_Defaults.pas
@@ -269,6 +269,13 @@ type
     cfmErase // Erasing plans
   );
 
+type
+  // Ware trade state
+  TWareTradeState = (
+    wts_Allow,    // Allow trade
+    wts_Default,  // Allow trade only if house, producing the resource was build
+    wts_Block);   // Block trade
+
 
 const
   MARKER_REVEAL = 1;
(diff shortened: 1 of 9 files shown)


Kromster80:
This change noticeably cuts Market use in the game.
E.g. players can no longer trade 1-2 horses for Scouts without investing into building a Stable.

reyandme:
Ok, but may be we can add this feature to MapEd, so mapmakers can decide to create maps for low PT games, where they use this option to not allow trade goods, if producing house is not build. By default resources are opened for trade.
Especially If I already implemented it, just few changes required...

Kromster80:
IIRC there's already "SetTradeAllowed" script for this.

reyandme:
Does it give same behaviour ? Its just allow trade it at some point. 
And I will forced to use script to get same behaviour, as a mapmaker. 

I think it will simplify life for some mapmakers without any changes on default gameplay. Just improvement.

Kromster80:
I don't think we need to make this new mode default, nor optional. Whoever needs it can easily make it using dynamic scripts.

lewinjh:
I agree. In missions where a resource is not available you can block trading that resource in the map editor. The market intentionally allows you to, for example, buy a few horses without a stables. And as Krom said you can unblock trading resources with dynamic scripts if that behaviour is needed for a certain map.

PR #147, opened 2016-12-06
Set default "center screen" position in the center of screen for new map, created in MapEd

reyandme:
instead of default (0;0) point.

Usually ppl will change it to a proper place, but its very usefull to have "center screen" point just in the center in case of creating many times small maps for testing purposes/building practice/etc, so after creating new map in MapEd (where any map opened in the center by default, which looks reasonable) you do not need to set its "center screen" position manually, so you can see what you have created in map ed immidiately after map start.


Kromster80:
Thanks!

PR #148, opened 2016-12-07
Ping/Fps label "shaking" when data changes, looks bad.

reyandme:
Idea:  we want to fix labels near slash ('/') symbol. Ping from the left, fps from the right, and slash sign always at same place without any "shaking" effect.
1 label is not possible for this, because font used there is not monospaced one, so we cant just Format text here. Label TextAlign taCenter is also not good (it was used before), because it use center of Label, instead of '/' symbol.
So I added 2 more labels for each player, because If we add only 1 label label ( label1='ping /' label2 = ' fps') then after player disconnection strikethrough effect looks not so good (it will affect slash symbol too, and at the labels junction strikesthrought effect looks bad). 

May be just make 2 separate columns for ping and fps ?


reyandme:
some gifs for "before-after" on trello card https://trello.com/c/vSEyjLVa

Kromster80:
Good work :-)
I'll merge when @home.

Kromster80:
Great, thanks :-)

PR #149, opened 2016-12-07
do not strike through team label, when player has no team

reyandme:
because them we will have team label caption as '-', do not need to strike it throught


Kromster80:
Do we have access to property that explicitly says that player has no team, instead of relying on "-" string?

reyandme:
Actually we do not want to strike through not because player has no team, but because we show it with '-' symbol. There is no reason to "strikethrough the strikethrough" itself :)
If later we will change that symbol (for example) we may be will want to strike it through again, if it looks ok.

PR #151, opened 2016-12-09
Highlight player flags in Lobby when they are clickable

reyandme:


lewinjh:
Looks good, but can you rename "DoShowPlayerMenu" to "CanShowPlayerMenu"?

"do" implies it will perform some action, "can" implies you are querying if it is allowed.

reyandme:
@lewinjh ok, fixed

lewinjh:
@reyandme thanks!

PR #153, opened 2016-12-13
Ability to have hints, dynamically changed on the same control objects

reyandme:
usefull when hint is changed by clicking on control, so mouse is still over the same obj and we need to check not only object, but also its Hint message. Used in "Chat mute player btn"


Kromster80:
Looks good to me. @lewinjh ?

lewinjh:
Yes looks good to me :)

PR #155, opened 2016-12-13
MapEd: House max out resources could be > 5

reyandme:
Depends of house production rate, max out res = 4+house prod rate
for quary and butchery: 7
for some others - 6


Kromster80:
Why do we need to complicate the code with this?

reyandme:
Hmm, I do not see any complications. 
Recently I tried to add 7 stones to quary in MapEd and realised that its not possible. I do not see why not

Kromster80:
Having more than MAX wares in a house is a workaround to keep produced wares. It was not intended to let houses MAX depend on house/ware type. It would make game more complicated by breaking the rule "each house fits MAX wares". There are no graphics drawn for this either.

PR #156, opened 2016-12-14
House closed for worker

reyandme:
new feature: ability to close house for worker. 
Behaviour: if house was occupied - leave it. If not - do not occupy.
UI: new button added for worker instead of icon: https://puu.sh/sOMEd/5e9944b51c.png


Kromster80:
Common 3D buttons toggle state and change picture in KM houses. I suggest to revert the changes to TKMButton3D and instead, use the TKMButtonFlat for worker icon/toggle.

reyandme:
after conversation on skype we will try to use 3d button, but with red line over the worker icon (probably line should go from left-top to right-bottom), instead of red triangle.

lewinjh:
Lets get this old PR finished so you don't have to keep merging it. The proposed solution (red line instead of triangle) sounds good to me.

reyandme:
#315 is opened as replacement of this PR

PR #157, opened 2016-12-14
Bugfix: Servername string corrupted in lobby

reyandme:
server sends us servername as UnicodeString, not AnsiString


Kromster80:
Thanks!

lewinjh:
This pull request was incorrect and reverted in PR #178. You can see in src/net/KM_NetServer.pas that mk_ServerName uses ANSI:

SendMessageA(NET_ADDRESS_ALL, mk_ServerName, aServerName);

I think the reason for restricting server names to ANSI was similar to restricting player names to ANSI. So everyone can read them even without Asian fonts enabled. Happy to discuss it if you want though.

PR #160, opened 2016-12-15
Save ware distribution between games

reyandme:
So you do not need to change it every game (very annoying).
Implementation:
- added new unit KM_WareDistribution.
- distributions saved locally in ini file packed, as 16 digits from 0 to 5.
- by default all distributions for all hands are created with DistributionDefaults values, so AI players always have same default ware distribution in every game
- when game starts new GIC command gic_WareDistributions initiated, which sends player current ware distributions (stored in settings) as packed string (same format, as it is stored in ini file). When game is loaded this command is not sended, because waredistributions are loaded from file

comes from our Buglist.


Kromster80:
@lewinjh Looks about ok to me. What do you say?

src/KM_Game.pas
```diff
@@ -443,6 +443,10 @@ procedure TKMGame.GameStart(aMissionFile, aGameName: UnicodeString; aCRC: Cardin
   if fGameMode in [gmMulti, gmMultiSpectate] then
     MultiplayerRig;
 
+  // Update our ware distributions from settings
+  if fGameMode in [gmSingle, gmMulti] then
+    GameInputProcess.CmdWareDistribution(gic_WareDistributions, gGameApp.GameSettings.WareDistribution.PackToStr);
+
```
lewinjh, line 451: I think it would be better to run this after we set the random seed and create the basesave. Just in case running this command uses the random number generator somehow in the future. Also the basesave is meant to be a save of the game state before we record any events through the GIP (so when watching the replay the basesave will have the default distributions and the first command replayed by the GIP will be to set the player's distributions). Basically running this command is like the player manually set all of the distributions, so we want to run it as late as possible so it's like the player did it.

lewinjh:
@Kromster80 Yes the feature and implementation both look good to me.

@reyandme Can you please have a look at my comment and fix the conflicts then we'll merge it.

PR #162, opened 2016-12-15
While spectating/watching replay change screen position when switch between players with dropbox.

reyandme:
Very usefull to move screen to see what is going on for this player instead of trying to find on Minimap (where sometimes its not easy at all)


lewinjh:
Won't think be an annoying feature though? Sometimes it's useful to look at the battlefield then switch between players to see which parts of fog are visible for each, without moving the screen position.

It should at least be optional, e.g. by holding CTRL while switching players it switches screen position, otherwise it does not move screen.

reyandme:
@lewinjh Ok, I think adding Ctrl hotkey for that will be nice. Because both options are very usefull (I mean switching to other player position immidiately, without trying to find him on the minimap, is usufull too)

reyandme:
@lewinjh We need to use Shift state in OnChange event for List and ColumnBox. We can add extra fOnChangeShift  TNotifyEventShift there, but then we have to manage witch OnChange (simple or with Shift state) to use, when creating ColumnBox. Otherwise we will invoke both methods, which is not good (even if it works fine, it is not expected behaviour), because we are setting fOnChange and fOnChangeShift here. 

So we can add flag in ColumnBox constructor, smth like aUseShiftStateOnChange: Boolean = False.
Or we can fully switch to TNotifyEventShift for OnChange method.

I prefered 2nd option - there nothing wrong to pass into the event handler method extra info, which could be or could be not used.

Actually I think its not good to have all these OnClick/OnClickShift/or even OnClickShiftPoint (with X,Y corrds). For instance recenty I wanted to do smth with doubleclick (select word in Memo/Edit on dblclick) - there was only OnDoubleClick with TNotifyEvent, but I wanted to use X, Y coords and ShiftState as well.  

So why its not good to pass extra info into event handler method:
1) method declaration (signature) is longer. I think its not a big deal
2) you need to pass extra 'fake' params into it (like nil, [] for shift state etc), when calling it manually. Its not a problem of event handler - you just do not need to call it manually, instead call another method, which do some logic, and also invoke it from event handling method. Event handlers should not be invoked manually. Sometimes we can invoke them ofc, but its not a good pattern.

@lewinjh What do you think ?

Kromster80:
@reyandme, @andreus791 I also remind you to make a single document/image listing ALL special shortcuts that are in the game now - all these special-case Ctrls and Shifts no one seems to know about.

lewinjh:
@Kromster80 I also dislike having so many hidden shortcuts that require holding CTRL or SHIFT. It would be nice if these are explained in the tooltip where possible e.g. "Feed group (hold CTRL to feed only soldiers that are hungry)"

An alternative would be to have a checkbox above the dropdown "Jump to selected player". But that's cluttering the overlay UI more.

@reyandme As Krom said you should not use events like OnDoubleClick internally within controls like Memo and Edit. If someone overrides that event then your selection behaviour will break. The On* events are intended for external use only.

reyandme:
I still think mass using smth like ReplayClick(nil) is a bad pattern. 
Also having both OnClick/OnClickShift looks strange. I want to get _outside_ some X, Y also, how can I get them ? Only add new OnClickPointShift.
And we have OnClickRight without Shift state (probably because when it was implemented, it was not needed outside). 
Don't you think something is wrong here ?

Anyway nobody wants to do some serious refactoring without a good reason.

What about Ctrl and shortcuts: added new PR #185 with all special shortcuts in new doc. 
And adding detailed info about all the shortcuts as hints is also planning, I have this task on Russian trello board.

Extra checkbox to "Jump to selected player" is not a good thing, as @lewinjh said, shortcut is much better.

Btw we were thinking about adding this hotkey: "Feed group (hold CTRL to feed only soldiers that are hungry)", but didn't conclude anything

lewinjh:
What happens if one of fLastSpecHandSelectedObj dies then you switch to that player? Or if the group has no members fMembers.Items[0] will crash.

IIRC gMySpectator.Selected has a way to handle this. Maybe fLastSpecHandSelectedObj should be in gMySpectator?

reyandme:
Yes, you are right, I missed that.
If we move fLastSpecHandSelectedObj[] to TKMSpectator then we could check all the conditions (not dead etc) in TKMSpectator.UpdateSelect method.

reyandme:
> Or if the group has no members fMembers.Items[0] will crash.

@lewinjh How can this happen ?
Anyway I added 
`and not ((aObject is TKMUnitGroup) and (TKMUnitGroup(aObject).Count = 0)) //We can not select groups with no warriors` 
for this

src/hands/KM_HandSpectator.pas
```diff
@@ -172,6 +209,18 @@ procedure TKMSpectator.UpdateSelect;
...
   if NewSelected <> nil then
     Selected := NewSelected;
+
+  // In a replay we want in-game statistics (and other things) to be shown for the owner of the last select object
+  if gGame.GameMode in [gmMultiSpectate, gmReplaySingle, gmReplayMulti] then
+  begin
+    UID := UID_NONE;
+    if Selected is TKMHouse      then begin HandIndex := TKMHouse    (Selected).Owner; UID := TKMHouse(Selected).UID; end;
+    if Selected is TKMUnit       then begin HandIndex := TKMUnit     (Selected).Owner; UID := TKMHouse(Selected).UID; end;
+    if Selected is TKMUnitGroup  then begin HandIndex := TKMUnitGroup(Selected).Owner; UID := TKMHouse(Selected).UID; end;
```
lewinjh: You forgot to change "TKMHouse" for the second two. Probably best to expand it rather than stacking several statements in one line, so mistakes like this are easier to spot

lewinjh:
Changes look good, storing UID makes sense.

Groups can have zero members after all members die if there are still active pointers to them. See:
```
//Group is dead, but still exists cos of pointers to it
function TKMUnitGroup.IsDead: Boolean;
begin
  Result := (Count = 0);
end;
```
Checking TKMUnitGroup(aObject).IsDead in your code would be neater.

reyandme:
@lewinjh fixed

Kromster80 (approved):

lewinjh:
@reyandme Thanks! (sorry for slow response)

PR #164, opened 2016-12-16
Save last selected items in lists

reyandme:
Task from our buglist.

Save in ini:
1) Replay viewer: SP or MP
2) Item in sp replay list
3) Item in mp replay list
4) Item in Load SP save list
5) Item in Start SP game list
6) MapEd: SP or MP
7) MapEd: item in SP list
8) MapEd: item in MP list
9) MapEd: new map sizes
10) MP lobby map type (fight/build/coop/special/save)

code refactored in related GUI units.

PeaceTimeLeft fix merged from andreus791:minox_fixes branch because game crashes if you try to view replay. He sent PR for that fix.


Kromster80:
How does this feature work - Is it intended to remember last selected items in all the lists and radios and set them on next game run?

I'm not sure this is a common design in other games/applications. When I ask to attach a file in Word/Excel, it remembers the folder, but it does not jump to last selected file.

reyandme:
Yes. It is intended to remember last selected items in all the lists and radios and set them on next game run.

reyandme:
I think next saving features are very usefull:
№ 1 Replay viewer: SP or MP
№ 6 MapEd: SP or MP
№ 10 MP lobby map type (fight/build/coop/special/save)

May be saving positions in list are not so important, but I think we still can add this possibility by some ini setting (one or for every separate list). 
For example when you creating new map, or do some tests, then you need to play one map many many times. Then its very annoying to scroll down to your map all the time. Even if its on the top - you need to select it. Even in windows sometimes I would like to have this feature.
I think we can add this feature optional for list positions as setting in the ini file. By default positions are not saved. 
And not optional for radios I mentioned above.

reyandme:
@Kromster80 I am still open for you to let me know what do you actually think ;)

Kromster80:
Okay, let's do this for radios and other "category" selectors. Keep the lists as they were.

reyandme:
After some skype conversations about this PR Krom stated, that this feature is not needed for lists, because of 'unnecessary complexity'. 

But I want to ensure you, that mainly this PR is refactoring all the code, what is related to these lists. And just minor peace of this PR is about new save list selection feature. 
There was at least 3 approaches how to handle list (in Load SP game, in MapEd and Replay), I unified them all to one approach, removed lots of duplicate code, removed invoking methods like List_Click(nil) etc.

Basically how does it work
1. When choose element in map/save list we save CRC of Map/Save and its name in the ini file to identify the map/save in future.
2. Then when open this list again we set fSelectedMap/Save CRC to saved one.

Let's look for instance at the references to saved CRC for MapEd SP list : 
references
Set and get only once. Same for all other CRC's. Very simple. If we do not need this I could simply set fSelectedMap to 0 (unselected) instead of saved one and do not save, when choose another item. Then this PR will just few code lines less for every list. 

Its very simple approach, there is no any complexity at all.

That is all - actual selecting is done by same mechanism like it was before: we go through the list and if fSelectedMap CRC is equal to current map CRC, then select it.

Of course refactoring can cause some errors and bugs. But we have to test them all, and that's it. And I tested it very well by now. We do not need to afraid of refactoring, if its a good one and provide better code maintenance.

So I believe it will be good to add this as an ini setting feature.

src/gui/pages_menu/KM_GUIMenuLoad.pas
```diff
@@ -159,19 +191,22 @@ procedure TKMMenuLoad.Load_Delete_Click(Sender: TObject);
...
+
+    Load_RefreshList(True);
   end;
 end;
 
 
 procedure TKMMenuLoad.Load_ScanUpdate(Sender: TObject);
 begin
-  Load_RefreshList(False); //Don't jump to selected with each scan update
+  Load_RefreshList(True); //Jump to selected with each scan update
```
lewinjh: I think the reason we previously didn't jump with each scan update was that when scanning lots of maps (especially on a slow computer) the list becomes unusable until the scan finishes because you can't scroll, it is constantly jumping back to the selected map. Is there a reason why you changed it?

There's a debug global variable to make scanning slower so you can simulate a first time scan (no cache) on a slow HDD.

src/gui/pages_menu/KM_GUIMenuLobby.pas
```diff
@@ -53,7 +53,8 @@   TKMMenuLobby = class
 
     procedure PlayersSetupChange(Sender: TObject);
     procedure MapColumnClick(aValue: Integer);
-    procedure MapTypeChange(Sender: TObject);
+    procedure MapTypeChange(Sender: TObject); overload;
+    procedure MapTypeChange; overload;
```
lewinjh: Why is this overloaded? What is the difference between the two versions? Presumably the Sender one is used when the radio is clicked, but then what's the other one for? It's not clear at all. If it serves a different purpose maybe it should have a different name.

lewinjh:
The refactoring looks ok, yes we were handling this inconsistently in different places. Why is it necessary to compare the save CRC and Name though? (we previously just compared CRC) We expect each save to have a unique CRC right? Checking the CRC only means you can rename saves on disk too.

Remembering which radio category you had selected (e.g. SP or MP replays) sounds good to me, since some players will mostly use one or the other.

As for remembering which item you had selected in lists, I don't have strong feelings about it. I can think of cases where it would be useful, like a map author opening KMR every day to edit the same map, now the map editor menu will have that map selected already.

src/gui/pages_menu/KM_GUIMenuLobby.pas
reyandme: I just wanted to get rid of MapTypeChange(nil) logic, because inside MapTypeChange there was an extra condition if Sender <> nil then do smth. 
I think its better not to use click handler as normal method, when passing nil to it, so its better to separate it into to different methods.
But overloading was not a good idea too. So I will rename these methods.

src/gui/pages_menu/KM_GUIMenuLoad.pas
reyandme: I checked it with debug variable and its very unpleasant. 
The reason why I set it to true is because I want to jump to selected after scan finished. For example on MapEd menu open or when switching map type.
So I think the proper solution will be jump to selected on scan complete.
I will refactor code due to this change

reyandme:
I've done some refactoring due to your review @lewinjh - renamed methods and refactored code about Jump To Selected item (now its jump to selected only on scan complete).

> Why is it necessary to compare the save CRC and Name though?

Well probably its not needed for save. Initially I added name in addition to CRC for Maps, because of rare occasion, when player download maps and rename it, then he could and download it again so we will have 2 identical maps with different names. And when sorting they could be selected randomly, depends of sorting order.

May be we can ignore this situations at all an rely just on CRC for both Maps and Saves. Or instead we can rely on Map/Save name, which is always unique.

> I can think of cases where it would be useful, like a map author opening KMR every day to edit the same map, now the map editor menu will have that map selected already.

This is the main scenario, why I implemented this feature. Its quite usefull in some situations.

reyandme:
Also I fix some small related bugs here, such as, when
1. there are two maps in maps list in MapEditor menu, 1st is from MapsMP and 2nd from MapsDL
2. You delete 1st map
3. Second map got selection
4. Buttons at the bottom of menu were not updated (so there was no button 'Move out of downloads')

src/gui/pages_menu/KM_GUIMenuLobby.pas
lewinjh: Great, thanks :)

lewinjh:
Fixes look fine.

@Kromster80 what do you think?

reyandme:
@lewinjh what do you think about this:

> Why is it necessary to compare the save CRC and Name though?
Well probably its not needed for save. Initially I added name in addition to CRC for Maps, because of rare occasion, when player download maps and rename it, then he could and download it again so we will have 2 identical maps with different names. And when sorting they could be selected randomly, depends of sorting order.

lewinjh:
@reyandme sounds ok to me.

Kromster80 (approved):

PR #166, opened 2016-12-17
Improved sorting of lists (maps/saves): sorting by MergeSort algorithm

reyandme:
For example when you sort maps by number of players there are plenty of maps with the same amount of players, and these maps are sorted randomly between each other. I fixed it introducing multi-parameter sort. So now in this case all maps with the same number of players will be sort by name.

Also changed sorting when sort by MapSize: 
before it was sorting by actual map size, which is not so important. Sort is mostly used to find the map you want, not to find the biggest map. And when you sort by it, the second sort parameter will almost never used (I set it to number of players). 
now I changed it so sort by size Label (XS/S/M/L/XL/XXL) (which also depends of actual map size, ofc), so second sort paramets is used, when labels are the same

Code refactoring in related GUI units.

**Update
Sort method changed to MergeSort. MergeSort applies in maps/saves lists and also in mp servers list


Kromster80:
Proper solution would be to switch to stable-sorting algorithm (e.g. MergeSort), so that items being sort on column A and then column B would preserve column A elements order when equal elements on column B are met. We already have a MergeSort implementation in KM_RenderPool. Not sure if it can be generalized to work in all places though from single source.

reyandme:
Added merge sort implementation. Works fine :)

lewinjh:
Mostly looks good.

It would be better if the map sizes are only declared in one place, so if we change them we don't need to find all the places to update. Maybe add a MapSizeIndex function which returns the index into your MAP_SIZES array, then make the function MapSizeText use MapSizeIndex and the MAP_SIZES array to return the text? (and the MAP_SIZES array can be moved into MapSizeText since that's the only place it will be used) Then in your sorting algorithm you can use MapSizeTextIndex to compare them rather than your IndexOf which is inefficient.

reyandme:
Ok. Well actually I added reference to MAP_SIZES array in MapSizeText function, so map sizes were declared in one place, but unfortunally lost this change when reverted initial changes.
But your solution is a little bit better, so I apply it )

src/KM_Utils.pas
```diff
@@ -199,21 +200,28 @@ function UTCToLocal(Input: TDateTime): TDateTime;
 {$ENDIF}
 
 
-function MapSizeText(X,Y: Word): UnicodeString;
+function MapSizeIndex(X, Y: Word): Byte;
 begin
   //Pretend these are understandable in any language
```
lewinjh: This comment is referring to the strings XS, S, M, etc. so should be moved there

PR #168, opened 2016-12-17
Minimap for MP games: show in replays menu and lobby, if possible

reyandme:
There was limitation for MP games minimap, they were not saving in *.sav file, because:
1) minimaps for different players are different, so *.sav files from different players will differ too. But when using save file for MP game we must be sure, that *.sav files are the same for every player to avoid inconsistency. Its very actual for "back to lobby" feature.
2) on minimap one player can possibly see smth about another player, what is hidden in game (in case full reveal minimap or minimap from possible enemy).

But minimap is very nice feature and it would be better if could have it in MP games too. For example when someone want to play again from save, he want to know, which save is better. So player could compare it by minimap - its not easy, but it can show approximately how big is your city. And it looks much better!

So I partially solved this problem by saving MP games minimap in a separate file with *.smm extension (Save MiniMap). Then problem number 1) is solved automatically, because your save file is same like before. Problem number 2) is also solved, because you can get minimap only from game you played (unless you manually put the minimap file, that somebody send you by skype, this option we obviously ignore). So you can not see anything new on that minimap.

Minimap is loading only if it possible, its not obligatory to have it.

Works well with "back to lobby".

The only thing that is not easy to fix - show minimap for new player (who did not play this save and do not have *.smm file). I think we can leave for now.

### DIFF
--- a/src/KM_Defaults.pas
+++ b/src/KM_Defaults.pas
@@ -218,6 +218,8 @@ const
   RETURN_TO_LOBBY_SAVE = 'paused';
   DOWNLOADED_LOBBY_SAVE = 'downloaded';
 
+  MP_MINIMAP_SAVE_EXT = 'smm';
+
 type
   TKMHandIndex = {type} ShortInt;
   TKMHandIndexArray = array of TKMHandIndex;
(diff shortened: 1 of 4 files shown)


lewinjh:
That's a neat solution :)

PR #169, opened 2016-12-17
House not occupied message should mention which house it is ("This farm is not occupied")

reyandme:
We need to add more TX_ labels for every house, because of plurals and fem/masc forms.

Task is from our buglist


src/houses/KM_Houses.pas
```diff
@@ -1305,7 +1305,7 @@ procedure TKMHouse.UpdateResRequest;
 
 
 procedure TKMHouse.UpdateState;
-//var HouseName: string;
+var HouseNotOccupMsgId: Integer;
```
Kromster80: It's better to rename "NotOccup" to "Unoccupied", so that the term used is consistent.

src/houses/KM_Houses.pas
```diff
@@ -1316,8 +1316,9 @@ procedure TKMHouse.UpdateState;
     Dec(fTimeSinceUnoccupiedReminder);
     if fTimeSinceUnoccupiedReminder = 0 then
     begin
-      //We can't paste houses name instead of %s like that because of plurals and feminine/masculine attrib
-      gGame.ShowMessage(mkHouse, TX_MSG_HOUSE_UNOCCUPIED, GetEntrance, fOwner);
+      HouseNotOccupMsgId := gRes.GetHouseNotOccupiedMsgId(fHouseType);
+      if HouseNotOccupMsgId <> -1 then // HouseNotOccupMsgId should never be -1
```
Kromster80: If it is -1, then maybe we need to log this?

src/res/KM_Resource.pas
```diff
@@ -64,6 +64,8 @@   TKMResource = class
     property Tileset: TKMTileset read fTileset;
     property UnitDat: TKMUnitDatCollection read fUnitDat;
 
+    function GetHouseNotOccupiedMsgId(aHouseType: THouseType): Integer;
```
Kromster80: It's better to rename "NotOccup" to "Unoccupied", so that the term used is consistent.

src/res/KM_Resource.pas
```diff
@@ -216,6 +218,37 @@ procedure TKMResource.LoadGameResources(aAlphaShadows: Boolean);
 end;
 
 
+function TKMResource.GetHouseNotOccupiedMsgId(aHouseType: THouseType): Integer;
+begin
+  case aHouseType of
```
Kromster80: This "case" is a bit superfluous, maybe we can do the usual house order trick?
`TX_MSG_HOUSE_UNOCCUPIED_ + Ord(HouseType)`

Kromster80 (changes requested):
That is a good idea to solve it like so. Please see about small improvement requests.

src/houses/KM_Houses.pas
reyandme: ok

src/res/KM_Resource.pas
reyandme: ok

src/res/KM_Resource.pas
reyandme: не для всех домов требуется такое сообщение, соответственно трюк будет не совсем такой:
сначала идет TX_MSG_HOUSE_UNOCCUPIED__22, для 1го дома (лесопилка), а затем уже 22 остальных дома, для которых это сообщение требуется. Но будет некая промежуточная зависимость между THouseType и теми домами, где требуется сообщение.

src/houses/KM_Houses.pas
reyandme: okay

PR #171, opened 2016-12-21
MP game stats - added info about citizens/soldiers died from starvation

reyandme:
Info added to existing bars (Citizens lost and Soldier lost).
Could be very usefull to see who of the players struggle of starvation in the city and who did not.
here how it looks like:
https://puu.sh/sWUSS/92ad599415.png

### DIFF
--- a/KM_TextIDs.inc
+++ b/KM_TextIDs.inc
@@ -276,9 +276,9 @@ TX_RESULTS_STATISTICS = 254; //Statistics
 TX_RESULTS_ECONOMY = 255; //Economy
 
 TX_RESULTS_MP_CITIZENS_TRAINED = 256; //Citizens|trained
-TX_RESULTS_MP_CITIZENS_LOST = 257; //Citizens|lost
+TX_RESULTS_MP_CITIZENS_LOST = 257; //Citizens lost|total/from|starvation
 TX_RESULTS_MP_SOLDIERS_EQUIPPED = 258; //Soldiers|equipped
-TX_RESULTS_MP_SOLDIERS_LOST = 259; //Soldiers|lost
+TX_RESULTS_MP_SOLDIERS_LOST = 259; //Soldiers lost|total/from|starvation
 TX_RESULTS_MP_SOLDIERS_DEFEATED = 260; //Soldiers|defeated
 TX_RESULTS_MP_BUILDINGS_CONSTRUCTED = 261; //Buildings|constructed
 TX_RESULTS_MP_BUILDINGS_LOST = 262; //Buildings|lost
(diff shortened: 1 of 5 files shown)


lewinjh:
I don't like the way it looks, it's too cluttered. What do you think @Kromster80?

Kromster80:
@lewinjh 
I'm not sure that "sum-up" should include that many details anyway. It is already quite hard to look at with 10 categories in two rows. Better not overload it with new info. Next we will want to add "friendly-fire" kills and kills by tower vs kills by men .. better keep it short and simple.

lewinjh:
@Kromster80 Yes I agree, it's meant to be a summary, lets keep is simple.

reyandme:
Then 'Units lost' or even 'Soldiers lost' summary is not clear. You could lose units from being invaded or just because your city has no food. And this scenarious are very different. 
Starvation is huge threat, you can't ignore it. One got great city, lots of soldiers and great weapons production could be almost defeated without anybody even attacked him because of hunger. 
I still think its very important to somehow let players know about it.

lewinjh:
I'm open to suggestions for how to visualise this, but I don't want to add more bars to the summary page or put two stats in one bar.

reyandme:
'Unit lost' is not clear how did they die. 

All other stats are very clear - trained units/built houses/production/kills/soldier equipped/destroyed buildings - all these stats are straightforward, easy to understand. 
Units lost - not easy at all, because it consists of 2 other stats, which origin is absolutely different. This statictics tells no so much now, just sum of 2 other measurements. Its X = Y+Z, where knowledge of X can't tell you something definite about Y and Z.

Btw stats of destroyed buildings are very unimportant info IMHO, usually 2-3-5 towers are destoyed and may be barracks/storage/school, and you can't say anything from that. I would better see how many units lost from hunger, then how many building was destroyed.

> I'm open to suggestions for how to visualise this, but I don't want to add more bars to the summary page or put two stats in one bar.

So that literally means no new stat could be placed on 1st page. What about other pages ? For instance Economy page. Graph is very long, it could be shortened and some info placed at the left side.

Or may be just add new page - 'Extra stats', which will looks like combination between 1st page (Statistics) and last (resources) page: 
on the right you have players niknames
on the left - list of different extra stats (or may be also main one, from the first page)
in the center area - bars for each player (different colors, best are somehow highlighted) for selected stat in the left list. Bars could be horizontal or vertical.
There we could add units and soldiers died from hunger/friendly fire/tower kills/etc etc (there were some other requests probably on forums for some specific staff).

Then do not put unit died from hunger into 'Unit lost' category on 1st page. Same for 'Soldiers lost'. 
Then this stat will be straightforward, easy to understand. 
Probably rename it then to something more precise, to let ppl know, that its not old 'Units lost' stats but smth different.

@lewinjh @Kromster80 what do you think about extra stats page? 
May be we can create an Issue for that ?

Kromster80:
New tab sounds okay.
It would be nice to allow to show there stats consisting of several params (google for stacked bars). E.g. [Unit losses] could show units lost from enemy, hunger, friendly fire, etc. Where each column would be captioned below with players name (so no need for right-side radiogroup).

reyandme:
Smth like this ?
Looks cool.

Kromster80:
Yes, much like this, with a list on left with all the extra stats players could need (presentable in bars form)

PR #173, opened 2016-12-22
Defeat player if he intentionally quit.

reyandme:
And notify other players about it with new message mk_Defeated.
So when no undefeated players left, except Player - he will receive Victory popup window.

Task is from our buglist


lewinjh:
Why is the mk_Defeated message necessary? Can't all players determine this by the fact the player quit?

And I'm not sure why we should mark goals as completed for defeated players? If the goal is "destroy enemy soldiers" why is this completed when the player is defeated?

Also, if the player is disconnected/times out (rather than intentionally quit), they should not be treated as defeated since they may rejoin the game using return to lobby (e.g. if their internet connection disconnected).

reyandme:
1) I'll refactor code about mk_Defeated. 
2) Well, I decided to use this to automatically popup victory window, which appears, when all goals are completed. 
Ok, we do not really complete some goals to win (for example, for multiplayer game these are default goals (to kill armies / destroy towns of all other players), but he should count as a winner even without it. It actually means, that when one player left the game, then other players goals about defeating him are not valid anymore.
So I can rename 'Completed' to 'Valid' or 'Enabled' (and reverse logic ofc). @lewinjh What do you think ?
3) Its not treated as defeated now, only if 'Quit mission' btn was clicked
4) Also there was an idea not to defeat player immidiately, give him some chance for, lets say, 2-3 minutes, so he could reconnect via Back to lobby mechanism. That happened some times for some emotional players, which can quit and then their friends can ask them to rejoin, so they could do that via back to lobby, not with autosave. This is actually Vovets idea, not sure is it really so important.

reyandme:
I've done code refactoring.
But there is still idea from my last message par.4.

Kromster80 (approved):
Idea sounds good. @lewinjh can you check the implementation?

lewinjh:
Sorry for taking a while to review this, I wanted to do it thoroughly because I could see problems. I agree with Krom the feature is a good idea. I've spent some time re-reading the code and thinking about it. There are some problems here which can cause bugs.

1. By setting the "Disabled" state on goals when a player is disconnected you are changing game state without going via the GIP, which is a bad idea. Each player could end up changing the game state on a different tick (because they are at a different game tick when mk_Disconnected arrives). This means the clients are out of sync. Player defeat can trigger script events and affect other aspects of the simulation so this is a problem. As a general rule everything owned by HandsCollection and Terrain is relied on to be in sync. So TKMNetworking should and the gameplay interface should never modified the game state.

2. You also change the game state unsafely when you call "gHands[gMySpectator.HandIndex].AI.Defeat(True);" from TKMGamePlayInterface.Menu_QuitMission (when the player quits the mission). On a side note, why is this necessary when the player is quitting anyway? (surely we don't care at that point)

3. fIsWinMessagePosted is not saved (also it's not synced so can't safely be saved), which means if you load a save where the game ended you will get a "team x wins" message immediately after loading.

4. Minor point but PostWinMessage and fIsWinMessagePosted are confusing names. We already have code that handles winning and displays the victory screen for players that won. What you have implemented is kind of an "end game" message when ALL players have won or lost. So maybe call it PostEndGameMessage instead of PostWinMessage?

5. Due to scripting and custom goals it's possible for multiple players to win who are not on the same team and your "end game" message doesn't handle that (there could be a special map where players all have different objectives).

We need to rethink the approach here. Is there a way to avoid changing the game state? Probably not since we want script defeat/victory events to fire when a player is defeated due to quitting. Maybe when a player quits the host sends a GIP command to defeat that player so the game state changes are synced. But what if the host quits, how do we ensure the new host sends the GIP command? (ideally not through more ugly special case variables in TKMNetworking like I complained about in #208). Then are there problems with delaying defeating the players via the GIP?

@reyandme what do you think? Happy to discuss ideas before you implement it.

reyandme:
Thank you for detailed review @lewinjh. I missed these issues.

1. I think we need to do as you suggested - use GIP command to be sure the game state is sync. To guarantee, that host will send it we can do these steps:
a. host receives mk_disconnected from player
b. host saves state that this player is disconnected by setting Dropped state to True. 
c. host sends gip command which defeats player by setting his goals as disabled and setting KM_Hand.AI.WonOrLost. 
Here we can't be sure that this command will be send to others because of maximum 0.1 sec delay for commands. But if it was sent, then all other players will receive it. 
So to ensure sending we must save somewhere the fact, that we handle it. Actually it is already stored in KM_Hand.AI.WonOrLost as game result. 
d. if host succesfully send the command, then all joiners will receive it and synchrosiously set him as defeated.
e. if host quits and do not send the command, then on mk_ReassignHost new host can check, is there any player with Dropped flag to True and still not defeated and not won hand (AI.WonOrLost = wol_None). I think this is enought to determine that rare occasion. I can't see how else player can be dropped and neither won nor lost.
f. then new host sends this command again (and if he quits - we can jump to chapter e. again)

Its a bit 'tricky', not 100% clear, but for me it's ok. We can add some method for this 'tricky' check: 'not dropped and not won and not lost'.

I think there is no problem of a little delay on defeating by GIP.

2. I do this just to set AI.WonOrLost  to wol_Lost, then player will see 'Defeated' title on the game stats page. We can use here GIP command as well. I think it will not be sent to anybody, because all the code is straightforward and just after sending GIP command there will be sending mk_Disconnected. But even if it not true, there is not problem if command will be sent 2 times. Or we can even check game state on the host side - using same check as for reassigned host.

3. so we have use another GIP here too ?

4. names of methods is not my best part for sure, probably because lack of english knowladge/skills. Ofc we can rename this method. Also we can say about all losers here too, same way - altogether in one line

5. Then it could happen that 1 player win at one moment, another at other and so on. Probably we can just show some local message when Hand.AI.Victory() happens.

lewinjh:
1. GIP command to defeat the dropped player sounds good. Your proposed solution for a reassigned host (checking for undefeated dropped players on mk_ReassignHost) also sounds good.

2. I don't like the idea of using the GIP command for the player that quit so they see DEFEATED on the result screen. Can't we just set PlayOnState := gr_Defeat in this case? No need to modify the game state. TKMGamePlayInterface.Menu_QuitMission uses PlayOnState to tell GameApp which message to show.

3. If you are using a GIP command for point 1 then this should now occur in sync, because it's that GIP command which will trigger fIsWinMessagePosted to be set. So you can safely save fIsWinMessagePosted to the save file?

5. Showing the message during player victory should be fine (then there's no need for fIsWinMessagePosted since player victory can only happen once). For the list of team members, maybe only display them if their state is also set to victory? (so if there's a scripted map where it's possible for only one member of a team to win, it doesn't list team members). But that might be tricky to handle since the victory state is updated one by one for hands, so when the first hand wins other hands will still be set to "not won not lost"...

Random thought: What should we do when the player times out and the host clicks "drop player from game" on the waiting screen? We probably shouldn't defeat that player since they might rejoin later once they fix their connection using back to lobby (or the "automatic reconnect after being dropped" feature we discussed).

lewinjh:
1. After thought: Checking for undefeated dropped players on mk_ReassignHost will cause problems when the host drops a player during a timeout, that player would be incorrectly defeated upon next mk_ReassignHost. Maybe we can just ignore this reassign case since it is extremely rare, and you can still defeat the player by destroying their town/soldiers.

reyandme:
1. Ok, so we can just ignore this rare case. Host should send new command on mk_Disconnect, which can happen only when player intentionally quit AFAIR

2. Does that mean that my current approach is OK then ? Because now in TKMGamePlayInterface.Menu_QuitMission only gHands[gMySpectator.HandIndex].AI.Defeat is called, which just set WOL status to defeat, post local message about it and show PlayMore dialog window

3. Ok. We will not need fIsWinMessagePosted if we use approach from bullet 4

4. So finally we can add victory message for every player, show it on Hand.AI.Victory. 
And EndGameMessage. 
EndGame message can happen when all the player won/lost. We can check this condition every 10 (lets say) ticks. 
But what to do with dropped players ? We do not need to defeat them as you said in your previous comment, but in case when there is no NOT DROPPED players with WOL = wol_None we can think the game is over and show End game message.

To be sure EndGame was at the same time for all players we can use GIP command here as well, and save defeated status in save too
And trigger new scripting function OnEndGame

EndGame message is going to be too complicated. Not sure we have to do it.

lewinjh:
1. Yes sounds good.

2. No, your current approach is unsafely modifying the game state when quit mission is clicked. I know the game is exited immediately afterwards, but it still seems like bad design to me. A better approach would be to pass gr_Defeat instead of gGame.PlayOnState. Take a look at TKMGamePlayInterface.Menu_QuitMission. Currently it calls gGameApp.Stop(gGame.PlayOnState); If you change that to:
`if (gGame.GameMode = gmMulti) and (gGame.PlayOnState = gr_Cancel) then`
`  gGameApp.Stop(gr_Defeat);`
`else`
`  gGameApp.Stop(gGame.PlayOnState);`
Then you will get the "DEFEATED" message when you quit a multiplayer game (if you haven't won already).
Displaying a local message about losing is pointless since when you quit the chat window is immediately destroyed.

3. Yes lets do it so fIsWinMessagePosted is not needed.

4. Rather than in Hand.AI.Victory you can use TKMGame.PlayerVictory for displaying the local victory message. That's the place where we play the victory sound and show the victory UI overlay, so it's the logical place to print the local message. By the way, why do we need to print a message about victory to the chat when you already see the UI overlay with "VICTORY!" written on it? (or "DEFEAT!" if you lost). An end game message seems unnecessary.

Why would we need a OnEndGame script event? The script can check victory/defeat status itself if necessary.

reyandme:
1. Ok, I''ll add new GIP command
2. Ok, I'll refactor according to your suggestion
3. Ok, I'll remove fIsWinMessagePosted
4. Ok, I'll refactor according to your suggestion

> Why would we need a OnEndGame script event? The script can check victory/defeat status itself if necessary.

No needed.

reyandme:
@lewinjh What about defeat delay ? 

From one of my previous messages:

> Also there was an idea not to defeat player immidiately, give him some chance for, lets say, 2-3 minutes, so he could reconnect via Back to lobby mechanism. That happened some times for some emotional players, which can quit and then their friends can ask them to rejoin, so they could do that via back to lobby, not with autosave. This is actually Vovets idea, not sure is it really so important.

Do we need this? 

If yes we can spread new GIP command into two commands: lets say gic_PlayerQuit and gic_PlayerDefeated. 1st command contains timestamp, which has to be saved in game. Then host check time in TKMNetworking.UpdateState and after 2-3 mins he sends 2nd command which actually defeats that player. 

If Back to lobby vote succeeded then defeat timer should be somehow paused probably. 
If player reconnected, then timer should be canceled.

reyandme:
@lewinjh could you please tell me your opinion on last question, so I will be able to complete implementation of this feature?

lewinjh:
No I don't think that's necessary. If the player purposefully quits (not accidentally lost connection) then that's their decision and they accept the consequences. Emotional players will probably do that less often after this change.

reyandme:
I agree.

reyandme:
I implemented this feature as we discussed earlier, with small changes:
- new GIP command gic_GamePlayerDefeat introduced
- fIsWinMessagePosted was removed as well as PostWinMessage
- I added win message in TKMGame.PlayerVictory. It will be shown not only to the winner (its not so important there), but also for other defeated/won (possible in scripts with multi winners) players who decided to stay in the game or for spectators.
- also on host reassign we need to send defeat command (gic_GamePlayerDefeat), because mk_disconnect from old host will come earlier, then mk_ReassignHost. Also want to mention, that continious sending gic_GamePlayerDefeat is possible without making any difference (its just will try to cancel all goals for this player again).

lewinjh:
> also on host reassign we need to send defeat command (gic_GamePlayerDefeat), because mk_disconnect from old host will come earlier, then mk_ReassignHost. Also want to mention, that continious sending gic_GamePlayerDefeat is possible without making any difference (its just will try to cancel all goals for this player again).

Host reassign isn't always an intentional disconnection, it could be that the host lost connection and will reconnect in a few seconds. So you can't just defeat them.

reyandme:
Ok.
Hmm... not sure what is better to do then when host quits... 
May be can check Dropped status on host reassign ? Dropped flag set to true means it was set on mk_Disconnect (from old host) as a joiner, before he become new Host.

src/KM_Game.pas
```diff
@@ -628,6 +631,18 @@ procedure TKMGame.GameMPReadyToPlay(Sender: TObject);
 end;
 
 
+procedure TKMGame.Game_OnReassignedHost(aDefeatedPlayerHandId: Integer);
+begin
+  OtherPlayerDisconnected(aDefeatedPlayerHandId);
```
lewinjh: As discussed this is not correct. Game_OnReassignedHost can be removed. Is fNetworking.OnReassignedHost unused then? Or does lobby bind it? The event can be removed if it's unused.

src/net/KM_Networking.pas
```diff
@@ -1684,7 +1720,8 @@ procedure TKMNetworking.PacketRecieve(aNetClient: TKMNetClient; aSenderIndex: In
                   //We are now the host
                   fNetPlayerKind := lpk_Host;
                   fMyIndex := fNetPlayers.NiknameToLocal(fMyNikname);
-                  if Assigned(fOnReassignedHost) then fOnReassignedHost(Self); //Lobby/game might need to know that we are now hosting
+                  if Assigned(fOnReassignedHost) then
+                    fOnReassignedHost(fNetPlayers[fHostIndex].HandIndex); //Lobby/game might need to know that we are now hosting
```
lewinjh: FYI we can't assume all players have a HandIndex, since spectators can be assigned to host. But since you are reverting the changes to fOnReassignedHost this isn't a problem (see my other comment)

lewinjh (changes requested):
This feature is far more complicated than I would have expected hahah. But we're getting closer to a working solution! I've had a full read of your changes again and made some comments.

Yes, the dropped flag is the correct way to tell when a player should be defeated. It means they either intentionally quit or they timed out and the host chose to drop them. Either way they should be defeated. Check everywhere that we call DropPlayer() during the game, we should be marking that player as defeated at that point (except for host disconnecting, see next paragraph).

Host reassigned is a special case. All of the joiners (non-hosts) call DropPlayer, but we can't send the GIC at that point since we don't know who the new host is. So yes, as you said check Dropped flag of the old host when the new host is reassigned, and send the GIC to defeat them then.

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -2649,7 +2653,7 @@ procedure TKMGamePlayInterface.AlliesOnPlayerSetup(Sender: TObject);
         and gGame.Networking.NetPlayers[NetI].IsHuman then // and is not Computer
       begin
         Update_Image_AlliesMute(Image_AlliesMute[I]);
-        Image_AlliesMute[I].Show;
+        Image_AlliesMute[I].Visible := True;
```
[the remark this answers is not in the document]
reyandme: well they should, but they are not ^) Show will *really* show control - it checks if parent is visible or not, if not, then show him too etc. Just settig visible will not change parent visible status. So next time Parent panel will become visible (when Allies panel is opened) it be shown as well. But we do not want to panel to appear after the line above.

This is not related to this PR actually.

reyandme:
Yes, this feature is not so obvious, that's true :)

I've done another refactoring, cleanup etc. One last thing:

> or they timed out and the host chose to drop them. Either way they should be defeated

Are you sure we have to defeat him? May be he will reconnect via back to lobby or new (possible) reconnect from MP list ?
F.e. to start back to lobby voting you have to drop old host, because menu controls are covered with Shape (of drop window PopUp) (ok, you can enter F4 menu by hotkey, but you can't go further). I believe its ok not to drop him then.
Current implemention works in this way - because first new host is reassigned, and only after it he can drop old host by pressing 'Drop player' button, but ReassignHost handler is already complited at that moment, so old host is not defeated afterall.

src/gui/KM_InterfaceGamePlay.pas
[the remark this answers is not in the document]
lewinjh: True, I forgot about that difference, thanks :)

lewinjh (approved):

lewinjh:
Ok you're right, if the host chose to drop the player they should not be defeated.

Finally merged! Sorry it took me so long to review it. I can't see any issues with the logic now but only thorough beta testing will tell ;)

src/KM_Game.pas
```diff
@@ -740,6 +753,12 @@ procedure TKMGame.RequestGameHold(Msg: TGameResultMsg);
 
 procedure TKMGame.PlayerVictory(aPlayerIndex: TKMHandIndex);
 begin
+  fNetworking.PostLocalMessage(Format('%s has won!', //Todo translate
+    [fNetworking.GetNetPlayerByHandIndex(aPlayerIndex).NiknameColoredU]), csSystem);
```
LauraRozier, line 747: Today we discovered that this line crashes the game when a player is victorious in SP.
Logic should check if fGameMode = gmMulti ( Could just move it down a little. ;) )

LauraRozier (commented):
There seems to be an issue with PlayerVictory that needs to be resolved.

src/KM_Game.pas
reyandme: I added 
```if IsMultiplayer then```
in my repository there. 
Moving down a little will cause not showing win message for spectators, so its better leave it here. Or refactor all of the method (for me its ok, not a big deal)

LauraRozier: Both sound OK to me, though I think @Kromster80 and @lewinjh might have some opinions on this.

PR #178, opened 2016-12-23
Dedicated server servername fix

reyandme:
Somehow we have to use UnicodeString when sending mk_ServerName ( I do know how that happened, that we were using AnsiString, which is ok, and then, suddenly, all servers posts there servernames as UnicodeString's. So we switched to UnicodeString as well in PR #157 . But then dedicated servers still sends server name as Ansistring, so this PR fixes that.

Also to compile dedicated server some Window Settings refactoring have been in another PR #177 , so I merged it to this PR too.

**Update
I found @lewinjh's commit from 18/12/2014 @ 2:26 PM, 'WelcomeMessage is Unicode (UTF-8 in INI file), ServerName is ANSI only'. Probably it was after release 6720. So it looks like we have to switch back to AnsiString for ServerName and then when all servers will use new dedicated server servername will look fine again (so do not merge this PR and revert PR #157 )
@lewinjh am I right ?

***Update 2
So finally a checked build with Server Name as AnsiString and same Dedicated server, and they works fine. So we need to revert changes from PR #157.

### DIFF
--- a/src/KM_Main.pas
+++ b/src/KM_Main.pas
@@ -94,6 +94,17 @@ end;
 
 
 procedure TKMMain.Start;
+  function GetScreenMonitors: TKMScreenMonitorList;
+  var I: Integer;
+  begin
+    SetLength(Result, Screen.MonitorCount);
+    for I := 0 to Screen.MonitorCount-1 do
+    begin
+      Result[I].Width := Screen.Monitors[I].Width;
+      Result[I].Height := Screen.Monitors[I].Height;
+    end;
+  end;
+
 begin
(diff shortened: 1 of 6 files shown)


lewinjh:
Ok, so the end result here is that PR #157 was incorrect and the only change in this PR is to revert #157? Seems correct to me, you can see in src/net/KM_NetServer.pas that mk_ServerName uses ANSI:

SendMessageA(NET_ADDRESS_ALL, mk_ServerName, aServerName);

So #157 was wrong.

I think the reason for restricting server names to ANSI was similar to restricting player names to ANSI. So everyone can read them even without Asian fonts enabled. Happy to discuss it if you want though.

PR #179, opened 2016-12-27
Game stats 'Weapons produced' now contains horses as well

reyandme:
added stats of warfare (which are weapons + horses). In MP stats its interesting to know not only weapons, but also horses, because they are actually used for war. To preserve weapons produced info I added them as extra info after '/' symbol.


lewinjh:
I don't really like this solution. Can we just include horses in weapons? We include shields/armour (which are not weapons) so why not horses? They are a ware that you produce like other weapons and they are used exclusively to equip soldiers. So IMO they should be counted as weapons for stats purposes.

If you need to see horses individually you can check the charts.

reyandme:
@lewinjh Ok, its fine for me add horses as weapons. 
Also we can rename them to 'warfare'. Not sure about is it a proper translation or not.

lewinjh:
Warfare is not a correct translation. It means the activity of war. I don't think English has a word that means military equipment and would be suitable in this context (medieval times rather than modern wars). So lets leave it as weapons for English, but other languages can use a more suitable word if one exists.

PR #180, opened 2016-12-28
MP chat focus improvement.

reyandme:
Game chat should not be in focus all the time. It must be possible to unset focus from without closing it, so any keyboard input processed as hotkeys, not as input typing in chat. When focus is unset it is still possible to read chat and using hotkeys.
To set focus - use 'open chat' hotkey, press button in game or just click on opened chat panel.
To unset focus - click anywhere outside of chat panel.

### DIFF
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -2384,6 +2384,9 @@ begin
 
   //We want these keys to be ignored by chat, so game shortcuts still work
   if Key in [VK_F1..VK_F12, VK_ESCAPE] then Result := False;
+
+  //Ctrl can be used as an escape character, e.g. CTRL+B places beacon while chat is open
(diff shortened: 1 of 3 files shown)


src/KM_Controls.pas
```diff
@@ -2730,10 +2736,6 @@ function TKMNumericEdit.KeyEventHandled(Key: Word; Shift: TShiftState): Boolean;
 
   //We want these keys to be ignored by chat, so game shortcuts still work
   if Key in [VK_F1..VK_F12, VK_ESCAPE] then Result := False;
-
-  //Ctrl can be used as an escape character, e.g. CTRL+B places beacon while chat is open
-  if ssCtrl in Shift then
-    Result := (Key in [Ord('C'), Ord('X'), Ord('V')]);
```
lewinjh, line 2736: Why did you remove these lines?

lewinjh:
Thanks

PR #182, opened 2016-12-29
Selectable text in edit and memo

reyandme:
By default text in all TKMEdit's and TKMMemo is selectable.
Added most of text editors selecting features.
In Memo added cursor for selection. Memo loses focus automatically when other controls are clicked


lewinjh:
Thanks! I always wanted to have this feature but was too lazy to implement it xD

PR #185, opened 2017-01-02
New doc for special shortcuts

reyandme:


Kromster80:
Let's move it into wiki or into Readme folder?

reyandme:
Readme's section of controls do not cover all of the controls, just main controls (no info about warriors controls). Also section about source codes is outdated.
Wiki have all the shortcuts, so I think its better to add this special shortcuts to wiki. But wiki is only in english.
Lets add it to both then :) But I am not sure my english is good enought for that, so I can add it for Russian readme and left English and other languages for somebody else. Currently we have Eng, Rus, Ger, Pol readme translations.
And now we can also change hotkeys, so we can mention it there too.

Kromster80:
Let's move it to wiki then =)

PR #187, opened 2017-01-04
Discussion PR. New ability to reconnect to the MP game from multiplayer page (server list).

reyandme:
This feature is common to many games and it will very nice to have it in KaM Remake too.

How should it work: 

MP game in Game stage. Player disconnects due to some network/pc problems (not intentionally quits). 

Now if this happens, then Back to lobby is the only chance for this player to come back again, or if other players will want to wait for him in the game (and his game did not crash). So his desire of reconnecting is up to host of the game (will he drop him or not) or other players (for back to lobby option).

How it should probably be:

1. After fixing his problems he will be able to reconnect to the game without going back to lobby and without game pausing.

2. In servers list his game is highlighted and always on 1st position in list. 
3. Player choose it and try to reconnect. 
4. Host should identify the player. For that purpose when game initially starts all the players get their UIDs, store them in ini file and use when reconnecting. Also do not allow players to change names, it will confuse other players. So identification should be done by pair of nikname and UID.

5. If identification was successful then game is paused, host make game save, player downloads it and reconnects to the game.
If identification was not successful, then error posted to player on multiplayer page. 

While downloading Host can drop player (in case of very bad player network connection).

Also there should be an ability to block abusers - who could all the time disconnecting via disabling network connection, then enabling it and trying to reconnect. May be Allies panel could have some icon near disconnected player name. 

Alternatively in par.5. we can show dialog window to the host, so he decides allow player to reconnect or not. Then abusers will not able to abuse anymore. 

Please share your thoughts about this feature.


lewinjh:
Rather than a pull request please log this as an issue and we can discuss it there. Yes I think it's a good feature and lets discuss it.

PR #196, opened 2017-01-14
Change port from string to Word ( Ends TODO in /src/net/KM_NetServerOverbyte.pas )

LauraRozier:
Signed-off-by: thibmo <thibmorozier@live.nl>

### DIFF
--- a/src/gui/pages_menu/KM_GUIMenuMultiplayer.pas
+++ b/src/gui/pages_menu/KM_GUIMenuMultiplayer.pas
@@ -6,7 +6,8 @@ uses
   {$IFDEF Unix} LCLType, {$ENDIF}
   StrUtils, SysUtils, KromUtils, KromOGLUtils, Math, Classes, Controls,
   KM_Controls, KM_Defaults, KM_Pics,
-  KM_InterfaceDefaults, KM_ServerQuery;
+  KM_InterfaceDefaults, KM_ServerQuery,
+  OverbyteIcsUtils;
 
 
 type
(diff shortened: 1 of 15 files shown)


Kromster80 (approved):

src/gui/pages_menu/KM_GUIMenuMultiplayer.pas
```diff
@@ -560,9 +566,14 @@ procedure TKMMenuMultiplayer.MP_ServersDoubleClick(Sender: TObject);
...
 
 procedure TKMMenuMultiplayer.MP_HostClick(Sender: TObject);
+var
+  serverPortStr: string;
+  serverPort: Word;
 begin
   //Save the player and IP name so it is not lost if something fails
   MP_SaveSettings;
+  serverPortStr := IcsTrim(Edit_MP_ServerPort.Text);
+  serverPort    := atoi(serverPortStr);
```
lewinjh, line 576: Is there a reason why you're using atoi instead of StrToInt? I didn't even know atoi existed in Pascal

LauraRozier: Both 'IcsTrim' and 'atoi' are introduced by the class 'OverbyteIcsUtils'.
It is also used by Ics, hence I used it as well.
I decided to follow their method to assure having the same result.

I think using StrToInt would be just fine.

Useful link : [atoi in OverbyteIcsUtils](https://github.com/Kromster80/kam_remake/blob/master/src/common/Overbyte%20ICS%208/OverbyteIcsUtils.pas#L2511)

lewinjh: Ok thanks for clarifying :)

Kromster80:
@thibmo Please look into fixing KaM_DedicatedServer and KaM_DedicatedServerGUI - they don't compile with these PR changes.

PR #198, opened 2017-01-23
Ability to ignore repeated consistency check errors while watching SP replay

reyandme:
'\n' in 642 text label replaced to EolW in TKMTextLibraryCommon.LoadLIBXFile

`s := StringReplace(s, '\n', EolW, [rfReplaceAll, rfIgnoreCase]); // EOL`


lewinjh:
This feature sounds ok to me.

I think we should include a warning that the replay will no longer be accurate and the game may crash, since a failed consistency check means the simulation will soon be very different to the original game and some command will eventually fail (e.g. attack a unit ID that was allocated to an ally instead of an enemy because the game randomness diverged). Right now it appears there will be no consequences to clicking yes to all, when in fact it means the replay has failed in a non-recoverable way.

Why does the LIBX English string say " (Todo translate)"?

reyandme:
I agree, its better to add warning. We can extend 642 text label with this warning message. May be you can write it ? Because my english is not so good.

I added "(Todo Translate)" to remember about translating this changed text label into other languages.

lewinjh:
In cases where the text has changed I usually delete the translations for all other languages to force the translator to fix it (downside is they have to completely rewrite it rather than just making changes). Our LIBX system doesn't have a way to indicate to translators that a piece of text is out of date and needs attention other than when there is no translation. Open to suggestions but the todo probably won't be seen by translators and might be left in the English translation by mistake.

Here you go:
"The replay failed a consistency check. This is a known bug which affects playback of some replays.\nIf you continue watching the replay will no longer be accurate and the game may crash.\nDo you want to continue watching? Choose 'Yes to all' to ignore future consistency check failures."

reyandme:
We can update translator to let him determine smth like "Todo translate" or just search in all of the files of the project by some tool for todo's.

Thx for warning msg.

Kromster80:
@lewinjh looks good to me. Please merge.

reyandme:
We can make new doc to remember all the texts we need to translate. We already have folder named 'Todo'... Hmmm

lewinjh:
@reyandme Good idea, can you start a document for this instead? I removed the todo from the text.

reyandme:
@lewinjh Ok, sure !

PR #200, opened 2017-01-23
Better scripting overlay placement depends of the game type 

reyandme:
Depends of the game type: game, spectate, replay 

During the game there is no replay/spectator controls and we can draw scripting overlay controls closer to the top of the screen. This will allow to have more free space, and overlay will not use too much of screen space.


lewinjh:
Good idea

PR #203, opened 2017-01-23
Replay/Spectator. Update screen position on DropBox List OnClick event, not only on DropBox OnChange event

reyandme:
So when you click onto already selected element screen position will be updated even if no OnChange event was triggered.

### DIFF
--- a/src/gui/KM_InterfaceGamePlay.pas
+++ b/src/gui/KM_InterfaceGamePlay.pas
@@ -1878,8 +1878,6 @@ begin
 
   if (Sender = Dropbox_ReplayFOW) then
   begin
-    gMySpectator.HandIndex := Dropbox_ReplayFOW.GetTag(Dropbox_ReplayFOW.ItemIndex);
-
     UpdateScreenPosition;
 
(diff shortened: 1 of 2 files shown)


src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -1843,27 +1880,7 @@ procedure TKMGamePlayInterface.ReplayClick(Sender: TObject);
   begin
     gMySpectator.HandIndex := Dropbox_ReplayFOW.GetTag(Dropbox_ReplayFOW.ItemIndex);
```
lewinjh: The line was duplicated in UpdateScreenPosition

reyandme: Yep, sure

lewinjh:
UpdateScreenPosition isn't a great name, it implies it would apply during normal gameplay not just replays (this is the main gameplay UI file). It's also not clear that it would change gMySpectator.HandIndex.

reyandme:
Yeah, that was not the best name, for sure. I missed that this is the main Interface gameplay file.
I've done some refactoring then.

PR #207, opened 2017-01-24
Show speedup clock when MP speedup for last player

reyandme:


src/KM_Game.pas
```diff
@@ -968,6 +967,14 @@ function TKMGame.IsMapEditor: Boolean;
 end;
 
 
+function TKMGame.IsMPGameSpeedUpAllowed: Boolean;
+begin
+  Result := (fGameMode in [gmMulti, gmMultiSpectate])
+        and (fNetworking.NetPlayers.GetConnectedCount = 1)
+        and not fWaitingForNetwork;
```
lewinjh: It would be better to check that all other human players have "Dropped" set in KM_NetPlayers rather than relying on fWaitingForNetworking during a disconnection/reconnection. Maybe add a function to KM_NetPlayers called OnlyOnePlayerRemaining (or a better name).

reyandme: Good idea.

reyandme:
I added new function to count not dropped players. Name is bit out of 'standart' (used negative form), but very easy to understand

Also couldn't stop myself from replacing 'i' to 'I' for iterators :)

lewinjh:
Thanks. Fixing iterators is fine, 'I' is the standard we agreed on :)

PR #212, opened 2017-01-25
Fix delphi compilation hints

reyandme:


src/gui/pages_menu/KM_GUIMenuMapEditor.pas
```diff
@@ -438,6 +438,7 @@ procedure TKMMenuMapEditor.ColumnClick(aValue: Integer);
 
 function TKMMenuMapEditor.GetMaps: TKMapsCollection;
 begin
+  Result := nil;
```
Kromster80: I came to a better pattern:

case of
  1:
  2:
else
  raise Exception.Create();
end;

It keeps compiler happy, takes less LOC and catches errors well.

Kromster80 (changes requested):
Can you use this pattern?

reyandme:
I find your solution is much better. I refactor all places where Assert(False) occurs.
And at least 4 times I find places where there was extra unused code. Compiler get more info about code when we use Exception.Create and can get some hints because of that.

There is still one last hint: unused private method:

`procedure TKMNetServer.SendMessageToRoom(aKind: TKMessageKind; aRoom: Integer; aParam: Integer);`

Can we delete it ?

lewinjh:
Yes it should be safe to delete SendMessageToRoom, I think it's a leftover from when we first implemented multiplayer. You only ever communicate to your own room, which is enforced by the server.

reyandme:
ok, I've deleted it

PR #214, opened 2017-01-26
Fix input ban all symbols except English

Alexandr57:
Small fix to ban entry of any simvlov other than English in the component MaskEdit - edtShortName


lewinjh:
We should probably allow numbers too

Alexandr57:
Hello lewinjh!
Like as in the Short Name Campaign from the beginning it was impossible to write numbers.
So even if I am permitted to introduce them in the code itself is MaskEdit not allow to do so.
PS: Sorry for my English. Translation with Google Translation

lewinjh:
Ok thanks :)

PR #215, opened 2017-01-26
Fix the position of nodes (red dots) in CampaignBuilder

Alexandr57:
Small fix problem solver output Nodes off-screen when you create.

### DIFF
--- a/Utils/Campaign builder/Unit1.pas
+++ b/Utils/Campaign builder/Unit1.pas
@@ -365,8 +365,6 @@ begin
     imgNodes[I].Top := Image1.Top + C.Maps[fSelectedMap].Nodes[I].Y - imgNodes[I].Height div 2;
     imgNodes[I].Left := EnsureRange(imgNodes[I].Left, Image1.Left, Image1.Left + 1024-imgNodes[I].Width);
     imgNodes[I].Top  := EnsureRange(imgNodes[I].Top, Image1.Top, Image1.Top + 768-imgNodes[I].Height);
-    //Nodes If the position is less than zero
-    //We are making it equal to zero
   end;
 
   shpBriefing.Top := Image1.Height - shpBriefing.Height;
--- a/Utils/Campaign builder/Unit1.pas
+++ b/Utils/Campaign builder/Unit1.pas
@@ -363,6 +363,12 @@ begin
     //Position node centers, so that if someone changes the nodes they still look correct
     imgNodes[I].Left := Image1.Left + C.Maps[fSelectedMap].Nodes[I].X - imgNodes[I].Width div 2;
     imgNodes[I].Top := Image1.Top + C.Maps[fSelectedMap].Nodes[I].Y - imgNodes[I].Height div 2;
+    //Nodes If the position is less than zero
+    //We are making it equal to zero
+    if imgNodes[I].Left < 0 then
+      imgNodes[I].Left := 0;
+    if imgNodes[I].Top < 0 then
+      imgNodes[I].Top := 0;
   end;
 
   shpBriefing.Top := Image1.Height - shpBriefing.Height;
--- a/Utils/Campaign builder/Unit1.pas
+++ b/Utils/Campaign builder/Unit1.pas
@@ -159,8 +159,8 @@ begin
     Img := TImage(Sender);
     Assert(Img <> nil);
 
-    Img.Left := EnsureRange(Img.Left + (X - PrevX), Image1.Left, Image1.Left + 1024);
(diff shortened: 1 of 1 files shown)


lewinjh:
Hi,
Can you explain it more detail please? How is it possible for the dots to be off screen? What about off screen in the positive direction (pos > width/height)?

And this is only enforcing it in the campaign builder UI, does that mean they are still saved to the campaign file with off screen coordinates?

reyandme:
I believe Alexander fixed default position for new red dots, here how it is now:

before
And after
after

Alexandr57:
> And this is only enforcing it in the campaign builder UI, does that mean they are still saved to the campaign file with off screen coordinates?

Hey.
No. I checked. In CampaignBuilder and in the remake of the same coordinates.
The bug lies in the fact that the Nodes (red dot) located at the center. And the initial coordinates were obtained - 5
imgNodes [I] .Left: = Image1.Left + C.Maps [fSelectedMap] .Nodes [I] .X - imgNodes [I] .Width div 2;
If mathematically and after checking through the debugger
imgNodes [I] .Left: = 0 + 0 - 10/2
Therefore the answer became -5
I just added a zero origin if they are negative.

Alexandr57:
Prior to the changes:
image
After the change:
image

lewinjh:
Thanks for explaining, that makes sense

Utils/Campaign builder/Unit1.pas
```diff
@@ -363,6 +363,12 @@ procedure TForm1.RefreshFlags;
     //Position node centers, so that if someone changes the nodes they still look correct
     imgNodes[I].Left := Image1.Left + C.Maps[fSelectedMap].Nodes[I].X - imgNodes[I].Width div 2;
     imgNodes[I].Top := Image1.Top + C.Maps[fSelectedMap].Nodes[I].Y - imgNodes[I].Height div 2;
+    //Nodes If the position is less than zero
+    //We are making it equal to zero
+    if imgNodes[I].Left < 0 then
+      imgNodes[I].Left := 0;
+    if imgNodes[I].Top < 0 then
+      imgNodes[I].Top := 0;
```
lewinjh: Maybe you should use EnsureRange here instead like we do in NodeMove? Otherwise the nodes could be positioned off the bottom/right edge of the image. And what about the flags, should we EnsureRange when positioning them in this procedure?

Also, a problem with this implementation is that the change does not get written back to the campaign data. So if the node is positioned at 0 in the campaign data, in the builder UI here we will set it to imgNodes[I].Width div 2 so it's not off screen, but when you save the campaign file will still have it at 0 and partially off screen.

Alexandr57:
I wrote to you that once placed nodes:
imgNodes [I] .Left: = Image1.Left + C.Maps [fSelectedMap] .Nodes [I] .X - imgNodes [I] .Width div 2;
     imgNodes [I] .Top: = Image1.Top + C.Maps [fSelectedMap] .Nodes [I] .Y - imgNodes [I] .Height div 2;
Returns - 5
On account of the 2
I noticed a bug. He keeps a - 5
Ie also maintains as the original

Alexandr57:
Problems with the record is not found. Writes right.
Cast Srinshot pliz
I just made corrections in mapping campaign editor but not in the remake

Alexandr57:
Problems with the record is not found. Writes right.
Cast Srinshot pliz
I just made corrections in mapping campaign editor but not in the remake
When recording in my correction:
C.Maps[fSelectedMap].Nodes[I].X and C.Maps[fSelectedMap].Nodes[I].Y
= 0

I tried this:
imgNodes[I].Left := EnsureRange(imgNodes[I].Left + (C.Maps[fSelectedMap].Nodes[I].X - PrevX), Image1.Left, Image1.Left + 1024-imgNodes[I].Width);
    imgNodes[I].Top := EnsureRange(imgNodes[I].Top  + (C.Maps[fSelectedMap].Nodes[I].Y - PrevY), Image1.Top, Image1.Top + 768-imgNodes[I].Height);
    imgNodes[I].Left := Image1.Left + C.Maps[fSelectedMap].Nodes[I].X - imgNodes[I].Width div 2;
    imgNodes[I].Top := Image1.Top + C.Maps[fSelectedMap].Nodes[I].Y - imgNodes[I].Height div 2;
But still in the end position in the Nodes editor campaigns = -5
image
image
What is the reason I do not know. But as the water Saves 0 and displays in the remake - 5

Alexandr57:
@lewinjh 
As you requested has changed:
`if imgNodes[I].Left < 0 then
imgNodes[I].Left := 0;
if imgNodes[I].Top < 0 then
imgNodes[I].Top := 0;`
on EnsureRange as you requested.
Hopefully now come.

PR #216, opened 2017-01-26
Update saved selection position in lists on save action

reyandme:
As continuation of PR #164, I expand usage of this new feature here.

In Map editor, on save map:
1. update saved position in map list on MapEd menu.
2. update radio btn MapType (SP/MP) on MapEd menu accordingly

In game, on save game:
1. update saved position in save list on Load game menu for SP games
2. update saved position in replay list on Replays menu
3. update radio btn ReplayType (SP/MP) on Replays menu accordingly

Then when player edit map, after save it will be his default choice
when player save game he will able to see his last game replay or load his last saved save easier.

Not sure if we need this for autosaves/
Also may be some players just never watch SP replays (which is reasonable, there no many things to watch), but sometimes play SP games, then they will be not happy with changed position of Replay type radio.

Then I can suggest to add ini setting for this rare occasion, to let ppl disable this 'autosave positions in lists' feature. 

P.S. also I added constants for maps/saves/campaigns/tutorials folders and replaced with them everywhere in the project


lewinjh:
Constants for map folders is good, thanks.

Auto selecting the last saved map/save for MapEd and singleplayer saves makes sense. I guess SP replays are ok too.

Autosave definitely should not change list positions, please confirm this isn't the case before I merge it.

Please lets not add INI settings for every possible tiny tweak like that. Otherwise we'll end up with 10 pages of settings that 99% of players don't use and don't understand what they do. Keep it simple.

reyandme:
> Autosave definitely should not change list positions, please confirm this isn't the case before I merge it.

It's not change the list positions now. Only manual save affects them.

Still not sure about replays. Most important saved position here is radio btn for Replay type - many of the players usually watch only one type of replay (mostly MP replay, I believe), then autochanging radio btn to SP replay type on SP game save could confuse them. 
So may we do not need update Replay positions on save.

Also there are a few more lists, where similar approach can be done:
1. Map list in Lobby
Probably we do not need it here, because its better to implement another suggestion from Buglist.txt - adding star for 'Favourite maps' (I've created new Issue #220 for this)
2. Save list in Lobby
Can be usefull to show last MP Save (even autosave ?) which is the most used case when players want to restart some game from some point. Not many players used in some other way - continue corporate game, or if they want to choose older save then the last one. Because saves are usually positioned one after another (similar name/creation time) then even If player wants to use older save, he could easier find it from last save. 
But still questionable. 
3. Campaign's list
It looks reasonable to continue playing last Campaign. So I think its good to add this feature here. And unset it (unselect campaign), when Campaign is finished, probably.

lewinjh:
1. The lobby should have nothing selected when you enter it as host, for example you might want to discuss with other players in chat before selecting anything.
2. Does not seem necessary.
3. Remembering last selected campaign sounds ok to me. Probably no need to unselect it when finished, you might want to play a specific mission again.

reyandme:
Ok. I will add same feature for Campaigns later

reyandme:
@lewinjh comment:
> I guess SP replays are ok too.

my comment:
> Still not sure about replays. Most important saved position here is radio btn for Replay type - many of the players usually watch only one type of replay (mostly MP replay, I believe), then autochanging radio btn to SP replay type on SP game save could confuse them.
So may we do not need update Replay positions on save.

@lewinjh What do you think about it ?

> Please lets not add INI settings for every possible tiny tweak like that. Otherwise we'll end up with 10 pages of settings that 99% of players don't use and don't understand what they do. Keep it simple.

Ok, sure

reyandme:
I removed update replay lists saved position on game save. Its not needed.

Added many string function in this PR to KM_Utils.pas, because I used some of them. These functions are needed for compilation on XE2. Also I will use them when implementing #247.

Unfortunally KM_Utils is also in linux lf endline format, so to see differences please merge #252 PR first.

src/KM_Game.pas
```diff
@@ -914,6 +915,20 @@ procedure TKMGame.SaveMapEditor(const aPathName: UnicodeString);
 
   fGameName := TruncateExt(ExtractFileName(aPathName));
 
+  // Update GameSettings for saved maps positions in list on MapEd menu
+  MapInfo := TKMapInfo.Create(aPathName, True); //Force recreate map CRC
```
lewinjh: Don't we need to free MapInfo when we are finished with it?

src/KM_Game.pas
```diff
@@ -1363,13 +1378,20 @@ procedure TKMGame.SaveGame(const aPathName: UnicodeString; aTimestamp: TDateTime
...
+  SaveInfo: TKMSaveInfo;
 begin
   //Convert name to full path+name
   fullPath := SaveName(aSaveName, 'sav', IsMultiplayer);
   minimapPath := SaveName(aSaveName, MP_MINIMAP_SAVE_EXT, IsMultiplayer);
 
   SaveGame(fullPath, aTimestamp, minimapPath);
 
+  // Update GameSettings for saved positions in lists of saves and replays
+  SaveInfo := TKMSaveInfo.Create(ExtractFilePath(fullPath), aSaveName);
```
lewinjh: Don't we need to free SaveInfo when we are finished with it?

src/KM_Maps.pas
```diff
@@ -1068,5 +1080,20 @@ procedure TTMapsCacheUpdater.ProcessMap(aPath: UnicodeString; aFolder: TMapFolde
...
+function GetMapFolder(aFolderName: UnicodeString): TMapFolder;
+var F: TMapFolder;
+begin
+  for F := Low(TMapFolder) to High(TMapFolder) do
+    if aFolderName = MAP_FOLDER[F] then
+    begin
+      Result := F;
+      Exit;
+    end;
+  raise Exception.Create('Unexpected map folder: ' + aFolderName);
```
lewinjh: Using Save from the F11 main menu might trigger this exception? It's not really unexpected to save outside of Maps and MapsMP.

lewinjh (changes requested):
Please address the comments. Otherwise looks good though.

reyandme:
Added fixes after review and conversation on skype

lewinjh (approved):

PR #221, opened 2017-01-28
Bugfix: side-effect assertion

### LINKED ISSUE #217: Compilation without assertions (-Sa)
If it is compiled without assertions, there is a minor bug in
https://github.com/Kromster80/kam_remake/blob/a443534/src/houses/KM_Houses.pas#L590
The result is that at least bowman's shoot at random positions all over the map.
The return value of the function call should be checked separately.

serztle:
This closes #217.


lewinjh:
Thanks

PR #222, opened 2017-01-30
Bugfix: do not allow empty filename when move map out of downloads or MapsMP folder can be deleted

reyandme:
1. There is some map in MapsDL folder
2. User press Move out of downloads button when this map is selected
3. clears map name edit field
4. press Move btn
Result - game will probably crash (sometimes it did not crash, idk why) and, what is much more important, **MapsMP folder will be DELETED**.

Probably it happens because when we move map we first try to delete existing map from MapsDL in case of same name (and overwrite checkbox checked), but if its name is empty then we delete parent dir...

MapsMP folder can be easily deleted with this 'trick' in r6720

in `procedure TKMapsCollection.MoveMap` only added first line with check `if Trim(aName) = '' then Exit;`, and fixed method indent.

### DIFF
--- a/src/KM_Maps.pas
+++ b/src/KM_Maps.pas
@@ -692,46 +692,48 @@ var
   Dest, RenamedFile: UnicodeString;
   SearchRec: TSearchRec;
 begin
-   Lock;
-   try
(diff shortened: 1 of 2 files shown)


lewinjh:
Thanks

PR #224, opened 2017-01-30
Campaign mission title

Alexandr57:
Add the ability to specify the title of the mission separately for each company rather than globally for all

Changes:
image
image
image
image

If the title of missions to be found will be displayed in the campaign libx file it if not that of a standard file

### DIFF
--- a/src/KM_Campaigns.pas
+++ b/src/KM_Campaigns.pas
@@ -54,6 +54,7 @@ type
 
     function CampaignTitle: UnicodeString;
     function CampaignDescription: UnicodeString;
+    function CampaignMissionTitle : UnicodeString;
     function MissionFile(aIndex: Byte): UnicodeString;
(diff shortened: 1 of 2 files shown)


lewinjh:
Why would campaigns need a different mission title? Isn't "Mission %d" fine for all of them?

Alexandr57:
lewinjh
Superfluous just will not.
For example if you want to kartodel in his campaign instead Mission 1 was Chapter 1, and not to change the main file.

Alexandr57:
Also, my method allows not show
Mission 1
and for example
Mission 1 of 10
Suffice it to libx campaign file write:
3:Mission %d of 10

So this feature is quite useful

Kromster80:
@lewinjh I guess it's a nice feature to have, as Alexadr shows in examples.

PR #226, opened 2017-01-31
Menu shortcuts (hotkeys)

reyandme:
Added shortcuts to all existing menu pages, code refactored, some controls improvements.

In details:
1. Added shortcut for going 'Back' in menu for most pages: ESC button. Not added for 'SP/MP Results' page to avoid occasional stats page leave. Work also with PopUp windows. Did not add shortcut to 'Exit game' to avoid unwanted exit.
2. For ColumnBox ENTER button triggers OnDblClick event handler. Then pressing ENTER will load save/open map in MapEd/load replay. I did not add this for 'Start game' in MP game for obvious reasons.
3. For PopUp windows (delete/rename/move) added ENTER handlers to confirm change. ESC = click on Cancel button.
4. Where possible added DELETE btn handler (delete map/replay/save game)
5. Refactor MapEd Delete/'Move out of download' and SP 'Load game' to use PopUp windows instead of buttons at the bottom of the list.
6. F5 button triggers Refresh server list on Multiplayer menu page
7. Added automatic focus cycling through any focusable control on every panel by pressing TAB.
8. Focus mechanism refactored a bit. Do not allow to focus on covered controls (under PopUp window, f.e.)
9. DropBox and DropList (with autoclose set to True) are now closable with ESC btn. 
10. Added HOME/END, PAGE UP/PAGE DOWN handlers for DropBox/ListBox. Better navigation for long lists (f.e. maps in MP)
11. Added shortcuts for 3 modal windows in MapEd for Ok (ENTER) and Cancel (ESC)

Current realisation do not play any sounds on shortcuts. Do we need to play them ?

It looks like we already have protection from spamming 'Refresh server list' querries, so holding F5 will not produce lots of traffic.


src/KM_Controls.pas
```diff
@@ -3154,7 +3154,7 @@ function TKMNumericEdit.KeyEventHandled(Key: Word; Shift: TShiftState): Boolean;
   end;
 
   //We want these keys to be ignored by chat, so game shortcuts still work
-  if Key in [VK_F1..VK_F12, VK_ESCAPE, VK_TAB] then Result := False;
+  if Key in [VK_F1..VK_F12, VK_ESCAPE, VK_RETURN, VK_TAB] then Result := False;
```
LauraRozier: ~~After a quick read :
Can you still send a chat message by pressing the return key? I'm quite sure that a lot of players enjoy that feature.
Should be tested to be sure. :-)~~
Numeric is not chat, never mind this comment. :-)

src/KM_Controls.pas
```diff
@@ -2491,8 +2612,8 @@ function TKMEdit.KeyEventHandled(Key: Word; Shift: TShiftState): Boolean;
     VK_END: Result := (fText <> ''); //These keys have no effect when text is blank
   end;
 
-  //We want these keys to be ignored by chat, so game shortcuts still work
-  if Key in [VK_F1..VK_F12, VK_ESCAPE] then Result := False;
+  //We want these keys to be ignored by TKMEdit
+  if Key in [VK_F1..VK_F12, VK_ESCAPE, VK_RETURN, VK_TAB] then Result := False;
```
LauraRozier, line 2672: After a quick read :
Can you still send a chat message by pressing the return key? I'm quite sure that a lot of players enjoy that feature.
Should be tested to be sure. :-)

reyandme: yeah, you can send chat messages with return key ofc, because it is handled in KM_GUIGameChat unit. I deleted this comment because we have many TKMEdits, which has nothing with chat, so its incorrect to point it here.
But what is a bit wrong with this code, that we need to return True from TKMEdit.KeyDown method, when return key was pressed and we did handle it (actually not only for return key, but for any handled key). This result (is key handled or not) should be passed from unit, where OnKeyDown is assigned (KM_GUIMessage in this case), because this is the place, where return key handled, and other TKMEdit may not handle it. Then OnKeyDown should be a function, not a procedure.
Currently KeyDown event is passed further, and could be handled in KM_InterfaceGamePlay also. This is working fine now, because there is no handlers for return key in KM_InterfaceGamePlay but could be a problem, when user change game hotkeys and set smth to return key.
I will do some refactoring due to this, Thanks!

reyandme: I have updated code because of this issue

LauraRozier: No problem. :-)

src/KM_Controls.pas
```diff
@@ -6427,8 +6709,21 @@ procedure TKMMasterControl.UpdateFocus(aSender: TKMControl);
...
+//Check if control is covered by other controls or not
+//We assume that control is covered if any of his 4 angles are covered
+//Use Self coordinates to check, because some controls can contain other controls (f.e. TKMNumericEdit)
+function TKMMasterControl.IsCtrlCovered(aCtrl: TKMControl): Boolean;
+begin
+  Result := (HitControl(aCtrl.SelfAbsLeft, aCtrl.SelfAbsTop) = aCtrl)
+        and (HitControl(aCtrl.SelfAbsLeft + aCtrl.SelfWidth - 1, aCtrl.SelfAbsTop) = aCtrl)
+        and (HitControl(aCtrl.SelfAbsLeft, aCtrl.SelfAbsTop + aCtrl.SelfHeight - 1) = aCtrl)
+        and (HitControl(aCtrl.SelfAbsLeft + aCtrl.SelfWidth - 1, aCtrl.SelfAbsTop + aCtrl.SelfHeight - 1) = aCtrl);
+  Result := not Result;
```
lewinjh: It would be less confusing to switch "=" with "<>" and "and" with "or" so you don't need to do Result := not Result

src/KM_Controls.pas
```diff
@@ -121,9 +127,15 @@   TKMControl = class
...
     procedure SetAnchors(aValue: TKMAnchorsSet); virtual;
+    function GetSelfAbsLeft: Integer; virtual;
+    function GetSelfAbsTop: Integer; virtual;
+    function GetSelfHeight: Integer; virtual;
+    function GetSelfWidth: Integer; virtual;
   public
     Hitable: Boolean; //Can this control be hit with the cursor?
     Focusable: Boolean; //Can this control have focus (e.g. TKMEdit sets this true)
+
+    IsMarkedToUnfocus: Boolean; //Used for cycle focusing on one TKMPanel
```
lewinjh: I don't really like this boolean, can we find a better way? You should be able to use TKMPanel.FocusedControlIndex instead. It looks like FocusedControlIndex is the index of the child which has IsMarkedToUnfocus set?

lewinjh (commented):
Overall sounds good, especially tab navigation and enter/esc to close popups.

I don't think ESC should exit the lobby to avoid accidental exits. It's an ok shortcut for closing popups and jumping back through unimportant menus (when you can easily return there), but not for important decisions like exiting lobby, stats, or game.

Don't you think the delete shortcut makes it too easy to accidentally delete maps/saves given that there's no undo or recycle bin? It's harder to click the delete button by mistake.

src/KM_Controls.pas
reyandme: ok

src/KM_Controls.pas
reyandme: Yes, it looks not obvious. But let me explain how it works. May be we will find neater solution.
Currently we have automatic focus mechanism, in TKMMasterControl.UpdateState function.
It is triggered whenever some element become visible/invisible/enabled/disabled/focusable/unfocusalbe (last 2 I added recently).

Before this PR, focus was chosen as the first element on specified panel, which satisfy focusable control restrictions: it should be enabled+visible+focusable. I also added here that he should not be covered with any other Control (even if control is enabled and visible, its possible to cover with it popup f.e.) 
Also I added another new restriction: if TKMPanel.FocusedControlIndex is set (<> -1) then we have to focus on it, so then focus could be not only on the first element in the list of appropriated panel controls, but on any of them. TKMPanel.FocusedControlIndex is set whenever focus happens 

Now lets see why IsMarkedToUnfocus (or similar mechanism) is needed: when press on Tab key we want to use the same TKMMasterControl.UpdateState function. And in TKMPanel.FocusNext we have to somehow indicate, that we do not want to have current focus anymore, but want to go to next control. That is why IsMarkedToUnfocus is here.

Probably neater solution could be to add to TKMPanel new function to get appropriate focusable control. And when Tab happens, we can get next one and set TKMPanel.FocusedControlIndex to a new value. Then in UpdateFocus we could also use same func (to do not spread not obvious code into 2 modules) and get new focused control relying on new TKMPanel.FocusedControlIndex.

reyandme:
> I don't think ESC should exit the lobby to avoid accidental exits. It's an ok shortcut for closing popups and jumping back through unimportant menus (when you can easily return there), but not for important decisions like exiting lobby, stats, or game.

I agree, I did not add shortcuts for Game Exit and Stats, but forgot about Lobby

> Don't you think the delete shortcut makes it too easy to accidentally delete maps/saves given that there's no undo or recycle bin? It's harder to click the delete button by mistake.

I agree here too. We have shortcut for Enter popup, so its enought. Delete shortcut is not needed

reyandme:
@lewinjh I've done some refactoring, as I mentioned above:
> Probably neater solution could be to add to TKMPanel new function to get appropriate focusable control. And when Tab happens, we can get next one and set TKMPanel.FocusedControlIndex to a new value. Then in UpdateFocus we could also use same func (to do not spread not obvious code into 2 modules) and get new focused control relying on new TKMPanel.FocusedControlIndex.

Also renamed TKMPanel fCollection to fMasterControl (probably fCollection is an old name due to old MasterControl class name)

lewinjh:
Thanks :)

PR #227, opened 2017-01-31
Scripting overlay small visual fixes/refactoring

reyandme:
- Fixed overlapping SpeedUp clock and overlay, when there was no ReplayFOW panel
- Added bevel for better text visibility. Bevel is not Hitable, so it ignores all mouse events

How it looks now:
1. No possible speed up clock and no ReplayFOW panel

2. Possible speed up clock and no ReplayFOW panel. 'Possible' means this ScriptingOverlay positioning will be same, even if clock is invisible  (when speedup = x1). This is done to prevent unpleasant overlay 'jumping' left and right when game speed changes

3. Possible speed up clock and also ReplayFOW panel

### DIFF
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -113,6 +113,7 @@ type
     function IsMapEditor: Boolean;
     function IsMultiplayer: Boolean;
     function IsReplay: Boolean;
+    function IsSpeedUpAllowed: Boolean;
     function IsMPGameSpeedUpAllowed: Boolean;
     procedure ShowMessage(aKind: TKMMessageKind; aTextID: Integer; aLoc: TKMPoint; aHandIndex: TKMHandIndex);
     procedure ShowMessageLocal(aKind: TKMMessageKind; aText: UnicodeString; aLoc: TKMPoint);
@@ -977,6 +978,12 @@ begin
 end;
 
 
+function TKMGame.IsSpeedUpAllowed: Boolean;
+begin
+  Result := not IsMultiplayer or IsMPGameSpeedUpAllowed;
+end;
+
+
 function TKMGame.IsMPGameSpeedUpAllowed: Boolean;
 begin
   Result := (fGameMode in [gmMulti, gmMultiSpectate])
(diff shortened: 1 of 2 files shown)


Alexandr57:
It looks like normal.

Kromster80:
Bevel - look too intrusive and "spoils" a lot of games screen area.
Positioning - Looks good. but I think it's better to always shift in one direction (Down).

reyandme:
1. Scriptmakers always complain about poor text visibility over the terrain. F.e. [sado's message on kam.net](http://www.knightsandmerchants.net/forum/viewtopic.php?f=32&t=2946#p47916) 
> -fixed color coding of barracks overlay, now hopefully more readable, still not perfect though...

Overlay 'spoils' game interface itself. 

Here is same text with and without bevel for 1024*768. With bevel:

without bevel

Without bevel:

with bevel

Except dark brown (and probably all dark colors) all other colors readability improves.
I barely can read gray, orange is also not easy, depends of background: word 'wooden' is easier to read, then 'shields' because of terrain shadow under word 'wooden'.

I tried to remove edge of Bevel, also different level's of shadow: We can compare all the result on one screen
No bevel:

with bevel

0.1 bevel

with bevel

0.2 bevel

with bevel

0.3 bevel

with bevel

0.4 bevel

with bevel

I like 0.3 or 0.2 dark level, without edge. Text reaability is much better, 'spoiling' is not significant.
Full screen with 0.2:

with bevel

2. I like shifting always in one direction, but think its better to avoid shifting down, because it uses too much of screen area (dividing screen area into two parts, above text overlay, and below it). Shifting to the right will use less space.
Probably its better to show overlay text on the right of +/- button to save a bit more space

Here how it look on 1024*768 with shifted down:

shifted down

I think overlay could be too low, in the middle of game viewport, that looks bad. For bigger resolution there is no such problem, but we can't rely on high resolution systems only.

P.S. Sorry for mass images, but there is no other way to show smth, then literally do it :)

Kromster80:
I'm against this PR. It spoils gameplay area a lot.

reyandme:
Ok, I removed Bevel.

> Positioning - Looks good. but I think it's better to always shift in one direction (Down).

Ok, I revert positioning as it was before (down)

Kromster80:
Thanks!

PR #228, opened 2017-02-01
Show cursor position in status bar for F11 debug menu

reyandme:


src/KM_FormMain.dfm
```diff
@@ -40,19 +40,24 @@ object FormMain: TFormMain
         Width = 110
       end
       item
-        Text = 'Cursor: 46.1 47.2'
-        Width = 160
+        Text = 'Cursor: '
```
Kromster80: Let's leave some placeholder numbers here, so we know what length of text to expect. Could be all 199.9 2999 1999 to indicate they are "fake".

Kromster80 (changes requested):

src/KM_FormMain.dfm
reyandme: ok

PR #229, opened 2017-02-01
Added ability to change log specification in F11 debug window

reyandme:
Also added ability to post logs into MP chat.

Added new group box for F11 debug window. Here it is:

Logs

- Delivery: turn On/Off delivery logs
- Net connection: turn On/Off network connection/reconnection/disconnection logs
- Net packets log level: radio group to define which net packets to log.
a. None: no packet will be logged
b. All but commands/ping/fps: all packets, except commands/ping/fps packets
c. All but ping/fps: all packets, except ping/fps packets
d. All packets
- Show logs in MP chat: all logs messages will be printed in chat, can be usefull sometimes, especially because we can copy into the buffer from chat window now. This feature replaces old global variable NET_SHOW_EACH_MSG. 

While in MP game if MULTIPLAYER_CHEATS var is set to False (by default for any release) then no network or delivery logs will be printed. This is made to avoid cheating because of possible log parsers.

### DIFF
--- a/src/KM_Utils.pas
+++ b/src/KM_Utils.pas
@@ -58,6 +58,8 @@ uses
 
   function GetNoColorMarkupText(aText: UnicodeString): UnicodeString;
 
+  function DeleteDoubleSpaces(aString: string): string;
+
   function GetMultiplicator(aShift: TShiftState): Word;
 
 implementation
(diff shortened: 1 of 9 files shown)


Kromster80:
Looks good to me. @lewinjh what do you say?

src/KM_Utils.pas
```diff
@@ -674,6 +676,23 @@ function GetNoColorMarkupText(aText: UnicodeString): UnicodeString;
...
+//Replace continious spaces with single space
+function DeleteDoubleSpaces(aString: string): string;
+var I: Integer;
+begin
+  Result := '';
+  for I := 1 to Length(aString) do
+  begin
+    if aString[I] = ' ' then
+    begin
+      if not (aString[I-1] = ' ') then
```
lewinjh: On the first iteration this could check aString[0] which is out of bounds and will crash. You should start the iteration from 2 instead.

lewinjh (changes requested):
Also looks good to me, just one minor issue I spotted

src/KM_Utils.pas
reyandme: sure

lewinjh (approved):

Issue #220, opened 2017-01-28
Ability to mark maps as a favourite (gold star)


reyandme:
From Buglist.txt:
Ability to mark maps as a favourite (gold star) so they appear at the top of the list (find your favourite maps without scrolling through whole list)

- - - - - - - - - - - - - - - - - - - -

Possible UI:
We can add new column with star in it for list in Lobby and Map Editor. For favorite map here is a star, for not favourite - star outline ('empty star'). Pics are clickable, so player can set/unset favourite status. 

We can even add kind of a rating here - 5 stars max, 0 min, then player could sort maps by number of stars. By default all maps are sorted by name and rating (first by name, then apply rating sort, then because of stable sort algorithm MergeSort, that we use now, same rated maps will be sort properly)

May be we can add same stars in MapEd menu where player also can set/unset them (it will need some place in MapEd menu to show these stars, but we can change controls placement on MapEd menu page, so minimap can be in the top left corner then map list will get enought width)

Implementation:
'Favorite' maps stored locally in ini file (by Name or CRC or both?) with their rating (probably we can add some format to add name/CRC/Rating value in 1 line, similar as I did for Ware Distribution). Then on maps scan new 'Favourite' property will be set for favorite maps. 

We already have gold star for Host, we can reuse it here. Then only new 'empty star' icon is needed.

thunder6:
Good afternoon,

I  have very similar problem with replay folder which need much more time to load the list. Sometimes need to wait approximately around a minute to be able to jump until the end. I dont have more files than 2697(divide it with 3= replays) in the replay folder(a bit more than 2GB). I use to save many games still what I spectated also.
It would be nice if this list could preload itself somehow in the backgroud when the game started up.

Idea2: Sometimes players want to share more than 2 or 3 replays with other players in the same time. In this case the only one downloaded file is not the best solution because the second downloaded file will overwrite the first one. could you restucturize the replay handling? I mean it could work as like the downloaded maps which have a second folder MapsMP and MapsDL so the replays would have also SaveMP and SaveDL or something like this.
MAny thanks!

Anyway the maplist is working nicely. ;)

Have a nice day!
t

reyandme:
Hi, thunder!

Interesting ideas, I think they worth implementing, but its better to discuss them in separate thread. Can you create 2 separate Issues for them ?

Thanks!

lewinjh:
[...]

PR #232, opened 2017-02-04
New feature - ability to mark map as favourite

reyandme:
Implementaion as discussed in #220


src/KM_Controls.pas
```diff
@@ -4927,6 +4957,36 @@ procedure TKMColumnBox.MouseMove(X,Y: Integer; Shift: TShiftState);
...
+  inherited;
+  MouseMove(X, Y, Shift);
+end;
+
+
+procedure TKMColumnBox.MouseMove(X,Y: Integer; Shift: TShiftState);
+begin
+  inherited;
+  UpdateMouseOverPosition(X, Y);
+  ProceedMouseDown(X, Y, Shift);
```
lewinjh: I don't really understand these changes. What does ProceedMouseDown do? It's only called from MouseMove so why is it called ProceedMouseDown?

lewinjh (commented):
Looks good overall

src/KM_Controls.pas
reyandme: Sorry, I misnamed this method. Think its better to put all ProceedMouseDown into MouseMove method

PR #235, opened 2017-02-08
MapEditor new feature: allow to grab an object with a MouseDown and drag it around until MouseUp 

reyandme:
Can drag units/group and houses.

Houses is a bit specail case, so I recorded a small video for this case:
http://youtu.be/8TwT-2mPa5s?hd=1


lewinjh (commented):
The feature is great, I feel like the implementation might be over complicated though. Maybe it would be simpler to always drag houses using the wireframe so you can avoid the erase road issue altogether and not have two different dragging modes for houses?

Also have you checked what happens if you move the mouse outside the game window or alt-tab while dragging? That can sometimes lead to strange behaviour with mouse events (e.g. drag is never released).

src/gui/KM_InterfaceMapEditor.pas
```diff
@@ -32,6 +32,14 @@   TKMapEdInterface = class (TKMUserInterfaceGame)
     fDragScrollingViewportPos: TKMPointF;
     fMouseDownOnMap: Boolean;
 
+    //Drag object feature fields
+    fIsDraggingObject: Boolean;           //Flag when drag object is happening
+    fDragObject: TObject;                 //Object to drag
+    fDragHouseGrabPntAdjustment: TKMPoint;  //Adjustment for house position, to let grab house with any of its points
+    fDragHouseEntranceTileOverlay: TTileOverlay; //Store tile overlay under house entrance, to avoid "road eraser" behaviour while dragging house
+
+    fStopMinimapUpdate: Boolean;  //Stop minimap from autoupdating, used in Drag House mechanism
```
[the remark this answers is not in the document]
reyandme: may be not so much

src/hands/KM_Hand.pas
```diff
@@ -121,7 +121,7 @@   TKMHand = class(TKMHandCommon)
     function CanAddFieldPlan(aLoc: TKMPoint; aFieldType: TFieldType): Boolean;
     function CanAddFakeFieldPlan(aLoc: TKMPoint; aFieldType: TFieldType): Boolean;
     function CanRemFakeFieldPlan(aLoc: TKMPoint; aFieldType: TFieldType): Boolean;
-    function CanAddHousePlan(aLoc: TKMPoint; aHouseType: THouseType): Boolean;
+    function CanAddHousePlan(aLoc: TKMPoint; aHouseType: THouseType; aConsiderEntranceOffset: Boolean = True): Boolean;
```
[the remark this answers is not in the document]
reyandme: ok, I'll check it

src/render/KM_RenderPool.pas
```diff
@@ -1356,7 +1356,7 @@ procedure TRenderPool.RenderForegroundUI;
                     RenderWireTile(P, $FFFFFF00) // Cyan quad
                   else
                     RenderSpriteOnTile(P, TC_BLOCK);       // Red X
-    cmHouses:     RenderWireHousePlan(P, THouseType(gGameCursor.Tag1)); // Cyan quads and red Xs
+    cmHouses:     RenderWireHousePlan(KMVectorSum(P, gGameCursor.CellAdjustment), THouseType(gGameCursor.Tag1)); // Cyan quads and red Xs
```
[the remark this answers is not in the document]
reyandme: I'll answer to this in comment down there

reyandme:
> Maybe it would be simpler to always drag houses using the wireframe so you can avoid the erase road issue altogether and not have two different dragging modes for houses?

May you are right. I have implemented this feature with these 2 types of house move because, I started from immidiate house move (same as for right click), then after tests realized how incovient it could be (with many houses/objects on the map you have no idea can this house fit into some area or not and actually why). So I added house markup. But probably having only house markup is good enought.

> Also have you checked what happens if you move the mouse outside the game window or alt-tab while dragging? That can sometimes lead to strange behaviour with mouse events (e.g. drag is never released).

I checked this situations:
- moving outside window does not make any problems, house is moved to the edge of the window, actually moved or not depends of map relief and objects
- when start drag object (or map, in DragScrolling feature) then alt+tab then release mouse btn, then alt+tab back to game window, then yes game thinks we are still in drag mode (object or scrolling), so f.e. any mouse move even without Holding mouse is fine. But then any Mouse Up event cancel it, as usual (for scrolling - its Middle btn mouse up only, for objects - any mouse up).

I am not sure now if we can handle alt+tab situation 100% clear, but we can add some protection from it: 
- on mouse move/mouse up check if proper mouse button is down in Shift state, so after alt-tab, if you do not have mouse down, drag mode will be reset.
- on any event (mouse down; key down/up - to avoid using hotkeys) except mouse up/mouse move - if we have DragObject (Scrolling) mode On, then reset it. 

About CellAdjustment:
Currently its reset, when *any* MouseUp event happens and we still have "DragObject" mode On (even on menu panel). 

But if we somehow change cursor mode avoiding mouseup event (by "2" hotkey, to open Houses page), then start f.e. build house somehow (future shortcut/using alt+tab, then mouse down+move any house) then yes we can get bug for a while.
But after protection I described above I do not see how we will be able to get it. We need sort of "finally" block - this (or other) protection should guarantee that.

Another approach (may be also nice one) is making gGameCursor a global variable of some TKMGameCursor class, then in Mode setter we will manage all "extra" states (CellAdjusment etc) and reset them if needed.

It would be interesting to know @Kromster80 opinion about this feature too.

Kromster80:
"Another approach (may be also nice one) is making gGameCursor a global variable of some TKMGameCursor class, then in Mode setter we will manage all "extra" states (CellAdjusment etc) and reset them if needed."

This is the way to go IMO. Just make sure TKMGameCursor known nothing about the gTerrain and such

reyandme:
@lewinjh @Kromster80 
I've done refactoring due to your suggestions.
- fStopMinimapUpdate removed
- aConsiderEntranceOffset removed
- CellAdjustment is managed in TKMGameCursor class, which replaced gGameCursor record
- alt-tab situations managed well (also for dragScrolling mode) with my suggestion:

> on mouse move check if proper mouse button is down in Shift state, so after alt-tab, if you do not have mouse down, drag mode will be reset.

src/gui/KM_InterfaceMapEditor.pas
```diff
@@ -32,6 +32,12 @@   TKMapEdInterface = class (TKMUserInterfaceGame)
     fDragScrollingViewportPos: TKMPointF;
     fMouseDownOnMap: Boolean;
 
+    //Drag object feature fields
+    fDraggingObject: Boolean;               //Flag when drag object is happening
+    fDragObject: TObject;                   //Object to drag
+    fDragHouseGrabPntAdjustment: TKMPoint;  //Adjustment for house position, to let grab house with any of its points
+    fDragHouseEntranceTileOverlay: TTileOverlay; //Store tile overlay under house entrance, to avoid "road eraser" behaviour while dragging house
```
lewinjh: fDragHouseEntranceTileOverlay can be removed now

src/gui/KM_InterfaceMapEditor.pas
```diff
@@ -900,6 +949,140 @@ procedure TKMapEdInterface.UpdateCursor(X, Y: Integer; Shift: TShiftState);
...
+
+
+//Drag house move mode end (with cursor mode cmHouse)
+procedure TKMapEdInterface.DragHouseModeEnd;
+var H: TKMHouse;
+begin
+  if (fDragObject is TKMHouse) then
+  begin
+    H := TKMHouse(fDragObject);
+    H.SetPosition(KMVectorSum(gGameCursor.Cell, fDragHouseGrabPntAdjustment));
```
lewinjh, line 987: Do we need to check if we are allowed to place the house here?

src/houses/KM_Houses.pas
```diff
@@ -465,6 +465,8 @@ procedure TKMHouse.DemolishHouse(aFrom: TKMHandIndex; IsSilent: Boolean = False)
 
 
 //Used by MapEditor
+//Return True, if house was set to new position successfully
+//aConsiderEntranceOffset - do consider house entrance offset while checking CanAddHousePlan and while set house to new position
```
lewinjh: Old comment about aConsiderEntranceOffset. Also it says it returns true but it's a procedure?

lewinjh (changes requested):
Recent changes look good, it's much simpler with just one house dragging mode

src/gui/KM_InterfaceMapEditor.pas
reyandme: Check is inside TKMHouse.SetPosition method already:
`if gMySpectator.Hand.CanAddHousePlan(aPos, HouseType) then`

lewinjh (approved):

PR #237, opened 2017-02-08
Map Editor: when select tile with eyedropper, scroll to this tile in tile table

reyandme:


lewinjh:
It looks like the new line format was changed in the files KM_InterfaceMapEditor.pas and KM_GUIMapEdTerrainTiles.pas. This makes diffs hard to read and messes up blame history. Could you please change them back to the format they were in before? (I believe we use Windows format) If we are not being consistent across our .pas files then please submit a different pull request to fix it.

reyandme:
After #236 I start facing this problems after merge from kromster80/master into my other branches with at least KM_InterfaceMapEditor.pas
I fixed all the PRs and now, after another merge it comes here again.

lewinjh:
That's annoying... It must have been switched somehow. The GitHub editor built into the website probably uses Linux style new lines, so maybe someone used that for a minor correction and it messed it up? Notepad++ lets you check and change it under Edit -> EOL Conversion.

reyandme:
#245 is going to fix cr lf for KM_InterfaceMapEditor.pas and KM_GUIMapEdTerrainTiles.pas

PR #239, opened 2017-02-08
Animation nodes in the Campaign

Alexandr57:
Animation for the nodes (red dots) in the Campaign


lewinjh:
The animation looks nice to me. @Kromster80 what do you think?

src/gui/pages_menu/KM_GUIMenuCampaign.pas
```diff
@@ -194,6 +199,26 @@ procedure TKMMenuCampaign.StartClick(Sender: TObject);
   gGameApp.NewCampaignMap(fCampaign, fMapIndex);
 end;
 
+procedure TKMMenuCampaign.AnimNodes(aTickCount: Cardinal);
+begin
+  if fCampaign <> nil then
```
Kromster80: this should be if fCampaign = nil then Exit;

src/gui/pages_menu/KM_GUIMenuCampaign.pas
```diff
@@ -194,6 +199,26 @@ procedure TKMMenuCampaign.StartClick(Sender: TObject);
...
+    if (aTickCount mod CAMP_NODE_ANIMATION_PERIOD) = 0 then
+      if InRange(fAnimNodeIndex, 0, fCampaign.Maps[fMapIndex].NodeCount-1) then
+      begin
+        if Image_CampaignSubNode[fAnimNodeIndex].Visible then Exit;
+        Image_CampaignSubNode[fAnimNodeIndex].Visible := true;
+        inc(fAnimNodeIndex);
+      end
+      else
+        Exit;
+  end else Exit;
```
Kromster80: Superfluous branch

Kromster80 (changes requested):

src/gui/pages_menu/KM_GUIMenuCampaign.pas
```diff
@@ -201,18 +201,14 @@ procedure TKMMenuCampaign.StartClick(Sender: TObject);
 
 procedure TKMMenuCampaign.AnimNodes(aTickCount: Cardinal);
 begin
-  if fCampaign <> nil then
+  if fCampaign = nil then Exit;
+  if Image_CampaignSubNode[fAnimNodeIndex].Visible then Exit;
```
lewinjh: If fCampaign.Maps[fMapIndex].NodeCount is equal to MAX_CAMP_NODES it will crash here after animating the final node

Alexandr57: I checked. Everything is working.
Animation is on the nodes on the map, not until recently.
For example, we have 10 nodes on the map. Consequently cycle will go from 0 to 9 instead of from 0 to MAX_CAMP_NODES

lewinjh: Did you test it with MAX_CAMP_NODES? (64) Think about it, the array Image_CampaignSubNode goes from 0 to 63. When fAnimNodeIndex=63 you make the final node visible (Image_CampaignSubNode[63].Visible := true). This is fine, but then we increment fAnimNodeIndex, so now it is 64. Next time AnimNodes nodes is called it checks "if Image_CampaignSubNode[64].Visible then Exit;". But Image_CampaignSubNode[64] is out of bounds so it will crash.

Alexandr57:
@lewinjh
fAnimNodeIndex will never be the same MAX_CAMP_NODES

lewinjh:
@Alexandr57 Why not? Looks like it will to me as I explained above

Alexandr57:
@lewinjh 
Fixed

Alexandr57:
@lewinjh 
Fixed

PR #240, opened 2017-02-08
Error PR. Map Editor: ability to set barracks rally point/woodcutters cutting point

reyandme:
Points can be set via 2 ways:
1. By clicking on button in house menu, then cursor changes to kmc_Beacon (do we need special cursor for this small feature?), then with left mouse button click on the map.
2. by clicking Shift + right mouse button on the map, when house (barracks/woodcutters) is selected. Its faster and also using right mouse button is more natural, because of using same button for this in the game.


reyandme:
Closed, due to some git craziness after merge (KM_InterfaceMapEditor.pas problems)

PR #241, opened 2017-02-08
Map Editor: ability to set barracks rally point/woodcutters cutting point 

reyandme:
Points can be set via 2 ways:
1. By clicking on button in house menu, then new marker appear (same as for reveal fow/center screen markers), then with left mouse button click on the map.
2. by clicking Shift + right mouse button on the map, when house (barracks/woodcutters) is selected. Its faster and also using right mouse button is more natural, because of using same button for this in the game.


src/houses/KM_HouseBarracks.pas
```diff
@@ -277,11 +279,24 @@ procedure TKMHouseBarracks.CreateRecruitInside(aIsMapEd: Boolean);
...
+  fRallyPoint := gTerrain.GetPassablePointWithinSegment(PointBelowEntrance, aRallyPoint, tpWalk);
+end;
+
+
 function TKMHouseBarracks.GetRallyPoint: TKMPoint;
 begin
   if not gTerrain.CheckPassability(fRallyPoint, tpWalk) then
     //update rally point if its not valid (not walkable). Set it closer to barracks entrance (PointBelowEntrance)
-    fRallyPoint := gTerrain.GetPassablePointWithinSegment(PointBelowEntrance, fRallyPoint, tpWalk);
+    SetRallyPoint(fRallyPoint);
```
lewinjh, line 308: Unrelated to this pull request (I only just noticed), but what you are doing here is unsafe for replays/multiplayer. GetRallyPoint should be side-effect free, since it will be called non-deterministically by the GUI when the player selects the house. This will cause multiplayer clients and replays to go out of sync when the previous rally point becomes unwalkable and the player selects the house.

Can you please submit a pull request to fix this? I suggest GetRallyPoint just returns fRallyPoint (so could be unwalkable), but then when you use  the rally point (when soldier is trained or woodcutter goes to work) you call ValidateRallyPoint first. This will be from within the game code so safe for replays/MP.

src/houses/KM_HouseBarracks.pas
```diff
@@ -277,11 +279,24 @@ procedure TKMHouseBarracks.CreateRecruitInside(aIsMapEd: Boolean);
 end;
 
 
+procedure TKMHouseBarracks.ValidateNUpdateRallyPoint;
```
lewinjh: I suggest calling this just "ValidateRallyPoint". Validate implies it will fix it if it is invalid, so I don't think Update is necessary. Same for woodcutter.

lewinjh (changes requested):
Also, new line format seems to have changed for KM_InterfaceMapEditor.pas in this PR too. Can you take a look at that please?

src/houses/KM_HouseBarracks.pas
reyandme: Ok, I 'll make new PR for this

lewinjh (approved):

reyandme:
I want to add to this PR small change: render cutting point where it will be set while mouse move. Please do not merge until that :)

reyandme:
Changes completed. PR ready to merge

PR #245, opened 2017-02-11
Fix for incorrect end line style (lf). Save in windows end line style (cr lf) instead

reyandme:
Only notepad - edit -> EOL converssion to win style has been done for these 2 files


reyandme:
@lewinjh can you apply this PR, so all my other PRs should become "readable" again

lewinjh:
Thanks!

lewinjh:
A few other files in your recent PRs have been swapped too, I'm guess it's when you merge changes from the master into PRs? It seems that Git assumes you use Linux style, and it can be set to keep your local copy using Windows style (but commit Linux style).

So the best solution might be to change everything to Linux style (so Github's editor/merge tool doesn't keep messing it up) and then set your preferences to checkout Windows style (that was an option that appeared for me when I installed Git).

@Kromster80 what do you think?

reyandme:
@lewinjh What other files have been "damaged" by setting to LF end line ?

It looks fine for now, isn't it ?

I am using Beyond Compare 4 to resolve conflicts, and GitKraken as Git client. Not sure if its important to know.

lewinjh:
There's still problems after this fix. In #241 KM_MapEditor.pas was swapped

PR #251, opened 2017-02-12
Fix: Rally/Cutting point: do not update in Getter, update manually when they used

reyandme:
Implementation of @lewinjh suggestion in #241:

> I suggest GetRallyPoint just returns fRallyPoint (so could be unwalkable), but then when you use the rally point (when soldier is trained or woodcutter goes to work) you call ValidateRallyPoint first. This will be from within the game code so safe for replays/MP.


lewinjh:
Thanks!

PR #255, opened 2017-02-13
Fix favourite maps inaccuracies

reyandme:
1. Refactoring. New class TKMFavoriteMaps introduced to encapsulate all logic related to favourite maps.
2. Fix maps sorting, after setting/unsetting map as favourite: now sorting will apply once on next list show
3. for drop down controls scroll position is now saved on Hide, and restored on Show (ItemIndex should be visible anyway). Actual for lobby maps drop down and for color select drop down. But work well for every frop down control
4. added sync farourite maps string in ini with current maps. Before if you change favourite MP map in MapEditor, its old CRC could stuck in favorite list (because nobody will never try to delete it by exact same CRC value) so after some time this line could become too long.

Also I wanted to add update favourite map CRC after its edit in Map editor. But for this it would be convinient to use PR #216. So lets wait untils #216 will be merged.


reyandme:
> Also I wanted to add update favourite map CRC after its edit in Map editor. But for this it would be convinient to use PR #216. So lets wait untils #216 will be merged.

I do add update favourite map CRC  after its edit in Map editor and resave with same name.

5. Also fixed: we need to sort with favourites only in lobby, do not do that in MapEd menu.

So this PR is ready for review.

Please approve #291 first, so it will be possible to see KM_Settings differences

src/KM_Controls.pas
```diff
@@ -5858,6 +5881,7 @@ procedure TKMDropColumns.ListChange(Sender: TObject);
 
 procedure TKMDropColumns.ListHide(Sender: TObject);
 begin
+  fListTopIndex := fList.TopIndex; //Save list scroll position
```
lewinjh, line 5887: Doesn't this need to be implemented for TKMDropList too? Currently TKMDropList.fListTopIndex is never set but is used in TKMDropList.ListShow

src/KM_Settings.pas
```diff
@@ -970,4 +944,122 @@ function TKMWindowParams.IsValid(aMonitorsInfo: TKMPointArray): Boolean;
...
+
+
+function TKMFavouriteMaps.PackToString: UnicodeString;
+begin
+  Result := fFavouriteMPMaps.DelimitedText;
+end;
+
+
+//Sync current Favourites Maps list with maps CRC array.
+procedure TKMFavouriteMaps.SyncWithMaps(aMapsCRCArray: TKMCardinalArray);
```
lewinjh: I find the name Sync confusing, it implies some kind of mirroring. Maybe RemoveNonExistentFavourites or something like that would be more descriptive?

src/gui/pages_menu/KM_GUIMenuLobby.pas
```diff
@@ -1744,7 +1776,7 @@ procedure TKMMenuLobby.MapColumnClick(aValue: Integer);
             SM := smBySizeDesc
           else
             SM := smBySizeAsc;
-      else SM := smByNameAsc;
+      else SM := smByNameDesc;
```
lewinjh: Why should the default sort be descending? That means it's Z..A by default?

src/gui/pages_menu/KM_GUIMenuLobby.pas
```diff
@@ -1718,6 +1739,17 @@ procedure TKMMenuLobby.RefreshSaveList(aJumpToSelected: Boolean);
 end;
 
 
+procedure TKMMenuLobby.MapList_OnShow(Sender: TObject);
+begin
+  if fMapsSortUpdateNeeded then
+  begin
+    //Update sort
+    MapColumnClick(0);
```
lewinjh: This will override the users previous sort next time they open the list. For example if I sort by map size, then mark a map as favourite, then close and reopen the list, the sort column will now be favourites not size (and Asc/Desc will flip each time this happens so sometimes favourites will end up on the bottom?). It would be better to trigger the sort again using the previously used sorting mode.

lewinjh (changes requested):

src/KM_Settings.pas
Kromster80: RemoveMissingFavourites would be shorter and more clear IMO (since we, humans, typically handle negations not easily).

lewinjh (approved):

PR #259, opened 2017-02-16
Added drag & drop function for Script Validator

Alexandr57:
Added drag & drop function for displaying the file name by moving it to the Script Validator

### DIFF
--- a/Utils/ScriptValidator/Unit1.pas
+++ b/Utils/ScriptValidator/Unit1.pas
@@ -1,9 +1,12 @@
 unit Unit1;
 interface
 uses
-  Classes, Controls, Dialogs, Forms, StdCtrls, SysUtils,
-  KM_Defaults, KM_Scripting;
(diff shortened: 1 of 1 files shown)


Utils/ScriptValidator/Unit1.pas
```diff
@@ -44,12 +45,15 @@ procedure TForm1.FormCreate(Sender: TObject);
   OpenDialog1.InitialDir := ExtractFilePath(Application.ExeName);
 
   fScripting := TKMScripting.Create(nil);
+
+  DragAcceptFiles(Form1.Handle, True);
```
Kromster80: Why do you address `Form1.Handle` instead of just `Handle` here?
Please change this and other places to just `Handle`

Kromster80 (changes requested):

Utils/ScriptValidator/Unit1.pas
Alexandr57: Ok

PR #260, opened 2017-02-16
Added KromUtils to the project

reyandme:
Its more convenient to see it in the project tree


Kromster80:
You can also add INC files to the project tree

PR #263, opened 2017-02-16
MapEd: add unit on mouse move with LMB. Fix issues with unit delete under cursor

reyandme:
1. while holding left mouse button with while in cm_Units mode will put Unit onto the map
2. Before delete unit used position of cursor over map cell, and when pointing to unit head unit was not highlighted with red. Especially for tall units (knight/scout). That is why it was a little bit unpredictable behaviour - to delete unit, you have to point to his legs.
3. added red highlight for unit groups when deleting


lewinjh:
Looks good thanks

PR #270, opened 2017-02-19
Remove Esc  btn hotkey for quit from lobby

reyandme:
I forgot to fix this in #226


lewinjh:
@reyandme thanks. Did you also remove the delete shortcut in the map editor/saves/replays menu? We decide it's too easy to delete given that there's no undo.

reyandme:
@lewinjh yeah, I fix it in #271

PR #272, opened 2017-02-19
Script additions

andreus791:
States.GroupOrdersBlocked, States.PlayerTeam, States.GetAllTeams

### DIFF
--- a/src/scripting/KM_ScriptingStates.pas
+++ b/src/scripting/KM_ScriptingStates.pas
@@ -24,12 +24,15 @@ type
 
     function GameTime: Cardinal;
 
+    function GetAllTeams: TIntegerArray;
+
     function GroupAt(aX, aY: Word): Integer;
     function GroupColumnCount(aGroupID: Integer): Integer;
     function GroupDead(aGroupID: Integer): Boolean;
     function GroupIdle(aGroupID: Integer): Boolean;
     function GroupMember(aGroupID, aMemberIndex: Integer): Integer;
     function GroupMemberCount(aGroupID: Integer): Integer;
+    function GroupOrdersBlocked(aGroupID: Integer): Boolean;
     function GroupOwner(aGroupID: Integer): Integer;
     function GroupType(aGroupID: Integer): Integer;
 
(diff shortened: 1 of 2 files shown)


Kromster80:
Please test these scripts in SP. It seems they might be crashing there

andreus791:
Tested, works well for MP but returns -1 for SP teams. Needs more work ;)

src/scripting/KM_ScriptingStates.pas
```diff
@@ -507,6 +511,31 @@ function TKMScriptStates.GameTime: Cardinal;
 end;
 
 
+//* Version: 7000+
+//* Returns an array with IDs of teams for all players
+//* Result: Array of team IDs
```
lewinjh: It's not clear what this function returns. Is it a mapping (lookup table) from Hand Index to Team? (that would make the most sense for the script). In that case you should iterate over hands count not NetPlayers count (so the length of the returned array is always hands count).

Otherwise it's unclear how this function could be useful to the script, since the indexes of the array will not match up with player indexes that the script uses in other functions (e.g. the player/hand index in Actions.OverlayTextSet would not be the same as the index into the array returned by this function).

src/scripting/KM_ScriptingStates.pas
```diff
@@ -507,6 +511,31 @@ function TKMScriptStates.GameTime: Cardinal;
...
+function TKMScriptStates.GetAllTeams: TIntegerArray;
+var
+  I, NetIndex: Integer;
+begin
+  try
+    SetLength(Result, 0);
+    for I := 1 to gGame.Networking.NetPlayers.Count do
+    begin
+      NetIndex := gGame.Networking.NetPlayers.PlayerIndexToLocal(I);
+      SetLength(Result, Length(Result) + 1);
```
lewinjh: No need to resize the array every loop (inefficient), just set the size once at the start.

lewinjh (changes requested):

Kromster80:
I agree with @lewinjh, we could accept the script commands to be "just for MP".
But first, it would be required to resolve #288, so we don't have to redo these script commands when Teams-Alliances resolution is found and implemented.

Kromster80:
**#288 needs to be resolved first**

lewinjh (approved):

PR #279, opened 2017-02-20
Batcher: new button to remove trailing empty comments in maps dat files

reyandme:


Kromster80:
I don't like RegExps where they can be avoided.
Why do you read maps through Parser.ReadMissionFileWOChanges  ?

Trailing spaces are not a big deal IMO. Trailing comments as well. Next thing we will trim double line-breaks and adjoin commands into one long line?

I would not want to reformat all maps "because we can", it serves no actual use.

PR #283, opened 2017-02-21
BugFix: close drop column/list when control become invisible

reyandme:
fixes this, when server was disconnected or somehow we switched page back while had opened drop list

bug

### DIFF
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -93,6 +93,7 @@ type
 
     fTimeOfLastClick: Cardinal; //Required to handle double-clicks
 
+    fOnChangeVisibility: TEvent;
     fOnClick: TNotifyEvent;
     fOnClickShift: TNotifyEventShift;
     fOnClickRight: TPointEvent;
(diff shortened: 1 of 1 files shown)


src/KM_Controls.pas
```diff
@@ -93,6 +93,7 @@   TKMControl = class
 
     fTimeOfLastClick: Cardinal; //Required to handle double-clicks
 
+    fOnChangeVisibility: TEvent;
```
lewinjh: IMO events should only be used by external things, not by subclasses. I would do it the following way:
 - Add UpdateVisibility as a virtual procedure of TKMControl (doesn't need to be implemented in TKMControl, it's just for subclasses)
 - Call it in TKMControl.SetVisibility
 - In TKMPanel override UpdateVisibility to call UpdateVisibility on all children (as it does now)
 - In TKMDropCommon override UpdateVisibility to close the list if the control is not visible (as it does now)

reyandme: Yes, I like your suggestion. It looks neater.

reyandme:
I've done refactoring due to your suggestion

src/KM_Controls.pas
```diff
@@ -171,7 +172,7 @@   TKMControl = class
     procedure Enable;
     procedure Disable;
     procedure Show;
-    procedure Hide;
+    procedure Hide; virtual;
```
lewinjh: virtual isn't needed now right?

reyandme: yep

PR #285, opened 2017-02-21
All maps update due to new batcher functions

reyandme:
As continuation of #142 and #278 
 
Used batcher with #284 fix

Next changes have been done for all maps *.dat files:
1. Removed all SET_NEW_REMAPs commands as unused 
2. For every player in all maps have add at SET_RGB_COLOR command where there is no one of commands SET_MAP_COLOR or SET_RGB_COLOR
3. Some maps have both SET_MAP_COLOR and SET_RGB_COLOR, SET_MAP_COLOR is always coming first, means SET_RGB_COLOR is always overriding color previously set by SET_MAP_COLOR. 4th button finds all such maps and deletes SET_MAP_COLOR command.


Kromster80:
Please do not adjoin separate things into one commit.
I have concerns about TPR maps (and some others), where REMAP is removed and MAP_COLOR remained, yet RGB_COLOR is set to be $FF****** instead of $******

reyandme:
> yet RGB_COLOR is set to be $FF****** instead of $******

fixed. Comment about maps changes added in PR descrition and new commit

PR #289, opened 2017-02-23
Lobby: for player drop box new possibility to set all slots as open/closed/ai

reyandme:
Added new column for drop list (actually its drop column now). 
'All' column should be small, to avoid unwanted clicks. To separate columns they use different colors for text. Also added highlighting for both columns with different colors.
Not sure about colors though.

I tested it with all languages and it looks fine for now. But even if some longer word will be used (not 'All', but 'Open all') then drawing should be fine, column box width is calculated and depends of translation.

May be instead of text ('All') we can add some icon. But probably short word is fine here.

Here how it looks now:

all


src/KM_Defaults.pas
```diff
@@ -674,6 +674,11 @@ interface
   icRed    = $FF0707FF;
 
   icSteelBlue = $FFA56D53; // Selection color
+  icRoyalYellow = $FF4AC7FF;
```
Kromster80: Please provide comments as to what these CONSTANT colors mean and where they are used and what for.

src/KM_Controls.pas
```diff
@@ -5339,13 +5369,16 @@ procedure TKMColumnBox.DoPaintLine(aIndex: Integer; X, Y: Integer; PaintWidth: I
       end else
       begin
         TextSize := gRes.Fonts[fFont].GetTextSize(Rows[aIndex].Cells[I].Caption);
-        if HighlightOnMouseOver and (csOver in State) and (fMouseOverRow = aIndex) then
-          Color := Rows[aIndex].Cells[I].HighlightColor //Brighten(Rows[aIndex].Cells[I].Color)
-        else
-        if HighlightError and (aIndex = ItemIndex) then
+        if not aDoNotHighlight
```
Kromster80: Can we avoid double negation here by changing the var name to aHighlight ?

Kromster80 (changes requested):

src/KM_Controls.pas
reyandme: aHighlight is not correct here, because we may highlight, or may not, depends on other reasons (that is why I used unnatural negative form, that was intentional).
Probably aAllowHighlight is Ok here

src/KM_Defaults.pas
reyandme: We decided to do some refactoring on skype, smth like
clRed = 
clPingHigh = clRed
clFpsLow = clRed

Here, I'll refactor, smth like:
icRoyalYellow = $FF4AC7FF
clChatSomething = icRoyalYellow;
clLobbyOpponentAll = icRoyalYellow;

reyandme:
We decided to do some refactoring on skype with @Kromster80 , smth like
clRed =
clPingHigh = clRed
clFpsLow = clRed

Here, I'll refactor, smth like:
icRoyalYellow = $FF4AC7FF
clChatSomething = icRoyalYellow;
clLobbyOpponentAll = icRoyalYellow;

Kromster80:
Looks good to me. I especially like the new Colors list. It looks tempting to move it to gRes.Colors now :-)

@lewinjh Please take a look and merge.

Kromster80 (approved):

PR #292, opened 2017-02-23
Selecting a folder for saving campaign instead SaveDialog + Fix Def Libx

Alexandr57:
Since the campaign name should be only info.cmp do not understand why do you need to use SaveDialog


Kromster80:
1. MapMaker should see into which file the campaign info gets saved.
2. MapMaker might want to save a copy of a info.cmp for different purposes (backup and etc.)

PR #295, opened 2017-02-25
Numbering Nodes and flags + Fix [Campaign builder]

Alexandr57:
Did numbering flags and nodes + Hints for flags and nodes + Fix

### DIFF
--- a/Utils/Campaign builder/Unit1.dfm
+++ b/Utils/Campaign builder/Unit1.dfm
@@ -131,19 +131,19 @@ object Form1: TForm1
     Top = 8
     Width = 513
     Height = 426
-    VertScrollBar.Position = 255
+    VertScrollBar.Position = 363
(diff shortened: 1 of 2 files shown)


Alexandr57:
```
shpBriefing.Top := Image1.Height - shpBriefing.Height - ScrollBox1.VertScrollBar.Position;
shpBriefing.Left := IfThen(C.Maps[fSelectedMap].TextPos = bcBottomRight, Image1.Width - shpBriefing.Width, 0) - ScrollBox1.HorzScrollBar.Position;
```
**This fixes:**
The RAD Studio XE 10 Here's a problem:

Due to the fact that if for example ScrollBox VertScrollBar.Position = 200 then Image1.Top = -200
image
image
Perhaps in other RAD Studio version of this bug no.
Please check your masters.

PR #296, opened 2017-02-25
Compiler warnings

reyandme:


Alexandr57:
The Delphi String is a limitation. It seems like one line of 255 characters.

lewinjh:
@Alexandr57 That's shortstring with the 255 limit.

We can't use AnsiStrings for filesystem stuff since files/folders can have unicode names.

PR #297, opened 2017-02-25
Fix: Neryn Campaign VAS04 mission crashes on open in MapEd

reyandme:
its because unit (serf) is on the same tile as other group. I moved that unit few to the tiles right.

Here is an error message:

error

Interesting, but mission starts well anyway :)


lewinjh:
Rather than crashing it should probably remove/skip the group and display a script warning, or find an alternative location for the group.

PR #298, opened 2017-02-25
BugFix: scripting action OnHousePlanRemove was invoked with wrong coord

reyandme:
Bug occurs, because when we invoke OnHousePlanRemove we use location, where user clicks (then GIP command is produced with click coordinates). Instead we must use HousePlan coordinates

### DIFF
--- a/src/KM_BuildList.pas
+++ b/src/KM_BuildList.pas
@@ -40,22 +40,26 @@ type
   end;
 
 
+  TKMHousePlan = record
+    HouseType: THouseType;
+    Loc: TKMPoint;
+    JobStatus: TJobStatus;
+    Worker: TKMUnit; //So we can tell Worker if plan is cancelled
+  end;
+
+
   //List of house plans and workers assigned to them
   TKMHousePlanList = class
   private
     fPlansCount: Integer;
-    fPlans: array of record
-      HouseType: THouseType;
-      Loc: TKMPoint;
-      JobStatus: TJobStatus;
-      Worker: TKMUnit; //So we can tell Worker if plan is cancelled
-    end;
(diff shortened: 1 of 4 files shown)


src/KM_BuildList.pas
```diff
@@ -782,19 +786,28 @@ procedure TKMHousePlanList.RemPlan(aLoc: TKMPoint);
...
+  Result.Worker := nil;
+end;
+
+
+function TKMHousePlanList.GetPlan(aLoc: TKMPoint): TKMHousePlan;
 var
   I: Integer;
 begin
-  Result := ht_None;
+  Result := GetEmptyPlan;
```
lewinjh: I would prefer the function to return a boolean for success and pass the plan via an out parameter rather than having GetEmptyPlan.

lewinjh (changes requested):
Good catch, I never considered that :)

src/KM_BuildList.pas
reyandme: ok

reyandme:
Well, I did not find this bug by myself, actually this bug comes from Toxic at [kam.net](http://www.knightsandmerchants.net/forum/viewtopic.php?f=23&t=3228#p48118)

lewinjh (approved):

lewinjh:
After reading the kam.net topic I'm wondering if we are consistent with house positions in script events/actions/states. Do we always use the entrance? (I believe all actions use the entrance, but some events/states might not) I think this event will give the offset position of the plan not the entrance...

@reyandme Could you check this please?

reyandme:
@lewinjh may be we can ask @andreus791, he knows (and uses) scripts better then me :)

PR #299, opened 2017-02-26
Change replay/spectator view on double click on players list

reyandme:
I tried our new feature to change viewport position by holding Ctrl and change/click on the player list.
But it looks much more natural to just double click on this list, then you moved to that player.

So I added this functionality. Not sure do we need Ctrl feature then. May be its still usefull for those, who will use keyboard's Up and Down in the players list. But doubleclicking is very nice.


lewinjh:
Yes double clicking seems more natural. Lets remove CTRL method, no need for duplicated functionality.

I just tested this and noticed a bug: If you select a group on one team, then a group on another team, then hold CTRL and click on the first team, the group of the second team is still shown in the interface on the left, even though the flag of the correct group is highlighted.

PR #303, opened 2017-02-26
Fix Position Nodes Campaigns

Alexandr57:
Fix Position Nodes Campaigns


Kromster80:
Please describe, what exactly did you fix?

Alexandr57:
Fixed position of the nodes for the animation nodes remake

PR #306, opened 2017-02-27
Replay view/Spectating: unfocus from players list on Esc key

reyandme:
Then this will let use arrow keys for map browsing.

Also moved to TKMControl fOnKeyUp/Down for easier maintanance of keys events globally in TKMControl, rather then in TKMEdit/List/ColumnBox etc.

Also added AutoFocusable flag to TKMControl, to mark if control can or can not be focused via our automatic focus logic in TKMMasterControl.UpdateFocus.
We can't use Focusable flag there, because we still have to be able to focus on control with mouse click by user.

This PR use #305 for testing convenience, so please merge #305 first.

### DIFF
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -16,7 +16,7 @@ type
   TNotifyEventKey = procedure(Sender: TObject; Key: Word) of object;
   TNotifyEventKeyFunc = function(Sender: TObject; Key: Word): Boolean of object;
   TNotifyEventKeyShift = procedure(Key: Word; Shift: TShiftState) of object;
-  TNotifyEventKeyShiftFunc = function(Key: Word; Shift: TShiftState): Boolean of object;
+  TNotifyEventKeyShiftFunc = function(Sender: TObject; Key: Word; Shift: TShiftState): Boolean of object;
   TNotifyEventXY = procedure(Sender: TObject; X, Y: Integer) of object;
 
   TKMControlState = (csDown, csFocus, csOver);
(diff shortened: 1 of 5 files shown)


src/KM_Controls.pas
```diff
@@ -4610,7 +4628,7 @@ function TKMListBox.KeyDown(Key: Word; Shift: TShiftState): Boolean;
                 end;
     else        begin
                   if Assigned(OnKeyDown) then
-                    Result := OnKeyDown(Key, Shift);
+                    Result := OnKeyDown(Self, Key, Shift);
```
lewinjh: Doesn't this mean fOnKeyDown will fire twice because we already called "inherited KeyDown" (which will call fOnKeyDown too)? Same for other controls like TKMColumnBox etc.

src/KM_Controls.pas
```diff
@@ -895,8 +900,8 @@   TKMColumnBox = class(TKMControl)
     HighlightError: Boolean;
     HighlightOnMouseOver: Boolean;
     Rows: array of TKMListRow; //Exposed to public since we need to edit sub-fields
-    OnKeyDown: TNotifyEventKeyShiftFunc;
-    OnKeyUp: TNotifyEventKeyShiftFunc;
+//    OnKeyDown: TNotifyEventKeyShiftFunc;
+//    OnKeyUp: TNotifyEventKeyShiftFunc;
```
lewinjh: Lets delete these

lewinjh (changes requested):

src/KM_Controls.pas
reyandme: Luckily it should not be invoked second time, because of
`if inherited KeyDown(Key, Shift) then Exit;`

But ofc its better to remove this. Same for TKMEdit.

And lets check all the controls then, to be sure

reyandme:
@lewinjh I checked TKMEdit/TKMMemo/TKMListBox/TKMColumnBox - they work fine, no double invokations

lewinjh (approved):

PR #308, opened 2017-02-27
Lobby: add all AI slots should be limited with available slots for selected map/save

reyandme:
For limit value we use:
- for save HumanCount value from save info
- for map locations count


src/KM_Controls.pas
```diff
@@ -905,6 +906,7 @@   TKMColumnBox = class(TKMControl)
     procedure SetColumns(aHeaderFont: TKMFont; aCaptions: array of string; aOffsets: array of Word);
     procedure AddItem(aItem: TKMListRow);
     procedure Clear;
+    procedure RevertLastItemIndexChange;
```
lewinjh: I don't like this function in the controls library, I don't think it will get used anywhere else, it's just to solve this specific issue. Can't we find a better way like not changing it to AI in the first place if the slot is not available?

lewinjh (changes requested):

src/KM_Controls.pas
reyandme: Yes, its kind of a "trick" to solve my issues. 
Why did I add it: ItemIndex of list is updated on every click on it - so when I click All for AI this DropColumn will be updated to ItemIndex = 2, even if should still be 0, because this ColumnBox is the last in list (f.e.) and there are not enought slots to set it to AI too, so I "revert" it.

The same thing could be done if we save this concrete ItemIndex not in the control but in Lobby class in some array or smth like that.

Also there are some issues, because of double usage of TKMColumnBox as separate control (servers list f.e.) and as DropDown item - behaviour is similar but there are some differences. May be its a good idea to split it into 2 controls in the future.

reyandme:
@lewinjh I refactored the code and put this special "revert" thing inside GUILobby class. Works fine !

src/gui/pages_menu/KM_GUIMenuLobby.pas
```diff
@@ -1133,7 +1143,18 @@ function TKMMenuLobby.SlotsAvailable: Byte;
...
 
 
-function TKMMenuLobby.DropColPlayersCellClick(Sender: TObject; const X, Y: Integer): Boolean;
+procedure TKMMenuLobby.DropBoxPlayers_Show(Sender: TObject);
+var DropBox: TKMDropColumns;
+begin
+  if Sender is TKMDropColumns then
+  begin
+    DropBox := TKMDropColumns(Sender);
+    fDropBoxPlayerSlot_ItemIndexes[DropBox.Tag] := DropBox.ItemIndex;
```
lewinjh: This seems kind of fragile, if we forget to update fDropBoxPlayerSlot_ItemIndexes somewhere. For example I'm not sure your solution will correctly handle when you select a save and remaining slots are set to closed/AIs automatically. Or when players join/leave the lobby and item indexes are updated.

Why can't we handle this problem completely in DropColPlayersCellClick? That would be much simpler than first applying the wrong settings and then undoing mistakes by keeping track of what the values were before we set them wrong?

I don't think we care about the case where the user already has the correct number of players, and sets an extra AI to "all" at the end. That's the user being silly and should happen very rarely. In that case it would be fine to let that slot be set to AI (but leave ones below it) and leave the user with one extra AI. Or set that slot to closed or open instead. Doesn't really matter because it should be very rare.

lewinjh (approved):

PR #312, opened 2017-02-27
SetField refactoring

reyandme:
1. Old SetField split into SetRoad and SetInitWine. everything related to corn and wine moved to SetFieldStaged
2. SetField removed
3. SetFieldStaged renamed to SetField.

### DIFF
--- a/src/terrain/KM_Terrain.pas
+++ b/src/terrain/KM_Terrain.pas
@@ -97,7 +97,7 @@ type
     procedure UnlockTile(aLoc: TKMPoint);
     procedure SetRoads(aList: TKMPointList; aOwner: TKMHandIndex; aUpdateWalkConnects: Boolean = True);
     procedure SetField(Loc: TKMPoint; aOwner: TKMHandIndex; aFieldType: TFieldType; aDoLocUpdates: Boolean = True);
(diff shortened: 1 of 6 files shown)


src/terrain/KM_Terrain.pas
```diff
@@ -1924,15 +1914,23 @@ procedure TKMTerrain.SetFieldStaged(Loc: TKMPoint; aOwner: TKMHandIndex; aFieldT
...
+  end;
 
   if (aFieldType = ft_Wine)
     and (InRange(aStage, 0, WINE_STAGES_COUNT - 1)) then
+  begin
+    if fMapEditor then
+      Land[Loc.Y,Loc.X].CornOrWine := 2;
+
+    Land[Loc.Y,Loc.X].Terrain  := 55;
+    Land[Loc.Y,Loc.X].Rotation := 0;
```
lewinjh: Looks like this will modify the terrain in MapEd which we don't want

lewinjh (changes requested):
Looks much neater this way, thanks

src/terrain/KM_Terrain.pas
reyandme: Thx! Forget about it

lewinjh (approved):

PR #313, opened 2017-02-27
MP game: change speedup default speed from 1 to game speed (before after PT). 

reyandme:
Used for toggle.

Also added support for x300 speed in MP, if allowed.

Also fixed small bug with checked x300 checkbox, when speed was changed from x300 to lower value by speedup hotkey.

### DIFF
--- a/src/KM_FormMain.pas
+++ b/src/KM_FormMain.pas
@@ -436,7 +436,6 @@ end;
 
 
 procedure TFormMain.chkSuperSpeedClick(Sender: TObject);
-var Speed: Single;
 begin
   if (gGameApp.Game = nil)
   or (gGameApp.Game.IsMultiplayer
@@ -445,12 +444,7 @@ begin
     and not gGameApp.Game.IsReplay) then
     Exit;
 
-  if gGameApp.Game.IsMPGameSpeedUpAllowed then
-    Speed := gGameApp.Game.GetDefaultMPGameSpeed
-  else
-    Speed := 1;
-
-  gGameApp.Game.SetGameSpeed(IfThen(chkSuperSpeed.Checked, 300, Speed), False);
+  gGameApp.Game.SetGameSpeed(IfThen(chkSuperSpeed.Checked, 300, gGameApp.Game.GetNormalGameSpeed), False);
 end;
 
 
--- a/src/KM_FormMain.pas
+++ b/src/KM_FormMain.pas
@@ -436,12 +436,21 @@ end;
 
 
 procedure TFormMain.chkSuperSpeedClick(Sender: TObject);
+var Speed: Single;
 begin
   if (gGameApp.Game = nil)
-  or (gGameApp.Game.IsMultiplayer and not MULTIPLAYER_SPEEDUP and not gGameApp.Game.IsReplay) then
+  or (gGameApp.Game.IsMultiplayer
+    and not gGameApp.Game.IsMPGameSpeedUpAllowed
+    and not MULTIPLAYER_SPEEDUP
+    and not gGameApp.Game.IsReplay) then
     Exit;
 
-  gGameApp.Game.SetGameSpeed(IfThen(chkSuperSpeed.Checked, 300, 1), False);
+  if gGameApp.Game.IsMPGameSpeedUpAllowed then
+    Speed := gGameApp.Game.GetDefaultMPGameSpeed
+  else
+    Speed := 1;
+
+  gGameApp.Game.SetGameSpeed(IfThen(chkSuperSpeed.Checked, 300, Speed), False);
 end;
 
 
--- a/src/KM_Game.pas
+++ b/src/KM_Game.pas
@@ -141,9 +141,7 @@ type
     property MissionMode: TKMissionMode read fMissionMode write fMissionMode;
     function GetNewUID: Integer;
     function GetNormalGameSpeed: Single;
-    function GetNormalMPGameSpeed: Single;
-    procedure SetNormalMPGameSpeed(aToggle: Boolean = False);
-    procedure SetGameSpeed(aSpeed: Single; aToggle: Boolean; aNormalSpeed: Single = 1);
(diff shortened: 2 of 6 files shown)


src/KM_FormMain.pas
```diff
@@ -436,12 +436,21 @@ procedure TFormMain.RGPlayerClick(Sender: TObject);
...
   if (gGameApp.Game = nil)
-  or (gGameApp.Game.IsMultiplayer and not MULTIPLAYER_SPEEDUP and not gGameApp.Game.IsReplay) then
+  or (gGameApp.Game.IsMultiplayer
+    and not gGameApp.Game.IsMPGameSpeedUpAllowed
+    and not MULTIPLAYER_SPEEDUP
+    and not gGameApp.Game.IsReplay) then
     Exit;
 
-  gGameApp.Game.SetGameSpeed(IfThen(chkSuperSpeed.Checked, 300, 1), False);
+  if gGameApp.Game.IsMPGameSpeedUpAllowed then
```
lewinjh: Why is this check needed? We should have already exited if MP speed up is not allowed. And even if it isn't, DefaultMPGameSpeed is what we should set the default speed to anyway right?

src/KM_Game.pas
```diff
@@ -1192,7 +1202,7 @@ procedure TKMGame.SetDefaultMPGameSpeed(aToggle: Boolean = False);
 end;
 
 
-procedure TKMGame.SetGameSpeed(aSpeed: Single; aToggle: Boolean);
+procedure TKMGame.SetGameSpeed(aSpeed: Single; aToggle: Boolean; aToggleSpeed: Single = 1);
```
lewinjh: How about we rename aToggleSpeed to aNormalSpeed?

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -2964,11 +2965,16 @@ procedure TKMGamePlayInterface.KeyUp(Key: Word; Shift: TShiftState);
     or gGame.IsMPGameSpeedUpAllowed
     or MULTIPLAYER_SPEEDUP then
   begin
+
     // Game speed/pause: available in multiplayer mode if the only player left in the game
-    if Key = gResKeys[SC_SPEEDUP_1].Key then gGame.SetGameSpeed(1, False);
-    if Key = gResKeys[SC_SPEEDUP_2].Key then gGame.SetGameSpeed(gGameApp.GameSettings.SpeedMedium, True);
-    if Key = gResKeys[SC_SPEEDUP_3].Key then gGame.SetGameSpeed(gGameApp.GameSettings.SpeedFast, True);
-    if Key = gResKeys[SC_SPEEDUP_4].Key then gGame.SetGameSpeed(gGameApp.GameSettings.SpeedVeryFast, True);
+    if gGame.IsMPGameSpeedUpAllowed then
```
lewinjh: Again I'm not sure why IsMPGameSpeedUpAllowed being false means the normal speed should be ToggleSpeed? If IsMPGameSpeedUpAllowed is false we shouldn't be changing speed anyway, but even so the normal speed for MP will be GetDefaultMPGameSpeed?

lewinjh (changes requested):

src/KM_FormMain.pas
reyandme: Default MP speed depends of game options and time (before/after) PT, while Default SP speed is always 1. Here we set normal speed to 1 or to gGameApp.Game.GetDefaultMPGameSpeed.
May be we can check gGameApp.Game.IsMultiplayer. Or better add new method which returns Default speed (for sp - 1, for mp - GetDefaultMPGameSpeed)

lewinjh: Ah I see, so you're basically using "IsMPGameSpeedUpAllowed" as "IsMultiplayer". Yes lets rename GetDefaultMPGameSpeed to GetDefaultGameSpeed and make it work with SP and MP to avoid these confusing checks.

src/KM_Game.pas
```diff
@@ -140,9 +140,10 @@   TKMGame = class
     property IsPaused: Boolean read fIsPaused write fIsPaused;
     property MissionMode: TKMissionMode read fMissionMode write fMissionMode;
     function GetNewUID: Integer;
-    function GetDefaultMPGameSpeed: Single;
-    procedure SetDefaultMPGameSpeed(aToggle: Boolean = False);
-    procedure SetGameSpeed(aSpeed: Single; aToggle: Boolean; aToggleSpeed: Single = 1);
+    function GetNormalGameSpeed: Single;
+    function GetNormalMPGameSpeed: Single;
+    procedure SetNormalMPGameSpeed(aToggle: Boolean = False);
+    procedure SetGameSpeed(aSpeed: Single; aToggle: Boolean; aNormalSpeed: Single = 1);
```
lewinjh: This is a lot of functions for handling game speed, lets simplify this:
1. Remove GetNormalMPGameSpeed. GetNormalGameSpeed will return MP speeds if it's MP, or 1 if it's SP.
2. Remove aNormalSpeed parameter. We can assume it's always GetNormalGameSpeed.
3. Remove SetNormalMPGameSpeed. It's only used in MultiplayerRig so lets just move those 4 lines there.

After that we should just have:
function GetNormalGameSpeed: Single; (only used inside TKMGame.SetGameSpeed now, but might be useful elsewhere later on)
procedure SetGameSpeed(aSpeed: Single; aToggle: Boolean);

Calls to SetGameSpeed are simpler because they don't care about what is the "normal" speed. That's TKMGame's responsibility.

lewinjh (approved):

src/KM_Game.pas
reyandme: Ok, that's better.
1. removed
2. removed
3. there is even 1 line after refactoring
4. GetNormalGameSpeed is used in 3 places are now

lewinjh:
Thanks, looks much cleaner now 👍

PR #315, opened 2017-02-27
House closed for worker feature (new PR)

reyandme:
GitKraken fatal error make crazy commit, so I make this PR instead of #156


reyandme:
Here how single red line looks like:

red line

src/units/KM_Units.pas
```diff
@@ -307,6 +313,79 @@ function TKMUnitCitizen.FindHome: Boolean;
...
+procedure TKMSettledUnit.CleanHousePointer(aFreeAndNilTask: Boolean = False);
+begin
+  if aFreeAndNilTask then
+    FreeAndNil(fUnitTask);
+  fHome.GetHasOwner := False;
+  gHands.CleanUpHousePointer(fHome);
+end;
+
+// Manage house is closed for worker process
+procedure TKMSettledUnit.ProceedHouseClosedForWorker;
```
lewinjh, line 325: I don't really like the way you've written this function, I think it's hard to understand and prone to edge case mistakes. It's also kind of deceptive, for example wIsInsideHouse can be true when the unit is eating inside the inn. Lets simplify it. I think there are bugs right now, for example:
wWorkingOutsideHouse can be true when the citizen is walking back into their house after working (while still visible). Then you free the task and nil their house which means they finish their walking inside action and are stuck homeless inside a house. There are probably other edge cases like this.

So how could we simplify it... Lets ignore walking in/out. Wait for that action to finish then handle it (otherwise we can end up with bugs like I mentioned above). This also means we can ignore going out to show hungry. 

So then cases are:
 - Unit is walking in/out: Wait for this action to finish
 - Throwing rock: As is, wait for it to finish
 - Going to eat: As is, reset fHome and leave task
 - Unit is idle/working inside the house: Stop working, clear home, walk outside
 - Unit is outside the house: Abandon walk, clear home, abort task

Not sure if this will make it much simpler or easier to understand... I think my main point is that rather than trying to handle every single case (which will be buggy and hard to maintain), we should wait for a good opportunity to abandon our task and clear our home (even if that means the woodcutter finishes planting his tree before he notices that he no longer has a home).

src/units/KM_Units.pas
```diff
@@ -307,6 +313,79 @@ function TKMUnitCitizen.FindHome: Boolean;
 end;
 
 
+procedure TKMSettledUnit.CleanHousePointer(aFreeAndNilTask: Boolean = False);
```
lewinjh, line 316: I don't like the name of this function. Maybe ResetHome would be better. I also think you should remove aFreeAndNilTask. That's unrelated to resetting the unit's home. It's only one extra line to call FreeAndNil(fUnitTask) and it's much more obvious what you are doing than a mysterious "true" parameter.

lewinjh (changes requested):
Looks good apart from ProceedHouseClosedForWorker. Sorry my comment is a bit rambling, I don't have a perfect answer for how to improve it but I don't like it as it is now. Happy to discuss it on Skype!

src/units/KM_Units.pas
reyandme: ok

src/units/KM_Units.pas
reyandme: Ok, I'll check it out, thx!
This feature was my very first for KaM Remake, so I probably missed something in October '16 :)

lewinjh: Hahah that's understandable. Our unit task/action code is quite fragile, you need to know about how all the tasks/actions work in lots of places. It's one of the oldest parts of the code base and not perfectly designed :)

PR #317, opened 2017-02-27
For multi monitor systems added new parameter to set border scroll area width on the second monitor

reyandme:
This PR fixes bug from our buglist:

> On 2 display setup, the game should not scroll the map if cursor is on the second display (windowed or fullscreen)

Difference is that you can set 2sn monitor scroll area width in ini for better game experience (without it its almost impossible to scroll with mouse to left or to right, depends of window/monitor position).

This case is not very popular, so adding this parameter in ini is good enought.

Currently we apply cursor restrictions when in Fullscreen mode, so this PR work only for window mode. After removing this restrictions in another PR it will work in fullscreen mode too.

I decided to add new parameter, because different users could expect different behaviour. 
One may want to have no restrictions at all (current and default behaviour) - most convinient way of mouse scroll on multi monitor systems.
Other may want to have full restrictions and use arrow keys for scrolling
And someone can choose something in between - f.e. if you set 100-200px, then it will be possible to scroll with mouse and also use your second monitor.

Could be usefull for streamers also.

### DIFF
--- a/src/KM_Main.pas
+++ b/src/KM_Main.pas
@@ -8,6 +8,8 @@ uses
   {$IFDEF USE_MAD_EXCEPT}, KM_Exceptions{$ENDIF};
 
 type
+  TRectArray = array of TRect;
+
   TKMMain = class
   private
     fFormMain: TFormMain;
@@ -46,6 +48,7 @@ type
     property FormMain: TFormMain read fFormMain;
 
     procedure ApplyCursorRestriction;
+    procedure GetMonitorsBounds(out MonitorsBounds: TRectArray);
     function GetScreenBounds(out Bounds: TRect): Boolean;
     function IsFormActive: Boolean;
     function ClientRect: TRect;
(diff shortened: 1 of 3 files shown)


lewinjh:
Obscure INI file settings are not user friendly for most people. Can we set this to something sensible? 10000 doesn't not sound like a sensible default.

> Currently we apply cursor restrictions when in Fullscreen mode, so this PR work only for window mode. After removing this restrictions in another PR it will work in fullscreen mode too.

Please don't remove this restriction. It's standard for games to lock the mouse to their monitor in full screen mode (you have to ALT-TAB out of a fullscreen game to use second monitor usually). If you want to interact with other windows at the same time use windowed mode.

reyandme:
> Obscure INI file settings are not user friendly for most people. Can we set this to something sensible? 10000 doesn't not sound like a sensible default.

I am not sure what is the highest possible resolution. I checked [this](https://www.w3schools.com/browsers/browsers_resolution_higher.asp) statistics, which has some strange line of "Other high resolutions	7.6%"

May be is better to set it to some default negative value (as off value) or just add another boolean switch on/off ?

> Please don't remove this restriction.

In #318 I added this as optional setting for reasons you mentioned, assuming anyway. that this adea presence in our Buglist means its worth implementing for some people. So I just implemented this task from Buglist. I have an idea why people would prefer fullscreen - there was Fraps/OBS cursor render bug in window mode before, on Fullscreen it was fine. I fixed it couple of months before. May be there some other reasons also why it was added to buglist., I do not know.

May be there some other unwanted lines in Buglist. Same happened when I implemented "Allow to block wares delivery into armor workshop" feature from there. 
Can you check our buglist to remove 100% unwanted/not actual features/bugs one day ?

reyandme:
Btw some games have extra options for "Cursor cage" (Pillars of Ethernity) or "Confine cursor" (Starcraft 2), in FullScreen and windowed mode. May be its a good idea to add this in Remake too (to options menu page). 
Some people want cursor cage (its much easier to scroll), but others need to use second monitor while playing, so they can use arrow keys for scroll.

Btw now I start thinking that idea I used in this PR with second monitor border area is not so good. May be its better to let to extend border area not to the second monitor, but widen it on the main one (with game on it). Now border area is 4px width. If one set it to 50px in ini file, than it could work for mouse scrolling and also allow to use second monitor.
@lewinjh @Kromster80 what do you think guys ?

**Update** I opened new Issue #322 about all this stuff

PR #323, opened 2017-02-28
Fixed position of nodes

Alexandr57:
Fixed position of nodes in all but The Federation of Neryn campaigns
In The Federation of Neryn I Do not know what should be the order of the nodes


Alexandr57:
@lewinjh 
Nodes are fixed. PR has been lying for a long time

lewinjh:
@Alexandr57 Please fix the nodes for Neryn, the order is clearly wrong now. Why don't you know the order to use? It looks clear to me, the nodes should move towards the selected mission. And the story makes it obvious anyway (e.g. in mission 4 it says you are going to an island).

Alexandr57:
@lewinjh There are some nodes there, so it's hard to determine which way the nodes should move. Can you fix it? I'll throw off the utility.

lewinjh:
Why is it hard to tell which way they should move? They should always move towards the selected mission.

Alexandr57:
@lewinjh 
Here you are talking from 3 to 4
Here is a screenshot:
image
There are no longer 3 to 4 but 3 to 5
And this is the lack of nodes:
image

Alexandr57:
Well, then what do you say?
image

Alexandr57:
Corrected in his opinion. The author of this campaign is not responding

Alexandr57:
Well, when will @lewinjh respond?
Has long been corrected.
The Neryn campaign is corrected without the author because the author does not respond.

LauraRozier:
It seems that Lewin is unable to respond at this moment.

I do see that multiple maps were changed, perhaps it would be a better idea to split them into separate PRs just in case of a change being wrong, then the maintainers can simple reverse the edit and fix it in a different way. :)

~~Let's have a check and see if the nodes are fine now. (For The Federation of Neryn)~~
I see a huge amount of missing nodes... Perhaps adding these might be nice. :)
Example:
Example

Alexandr57:
@thibmo I can not do this smoothly.

LauraRozier:
I did add some quick nodes, looks bad but it's OK.
Now for adding new nodes.. Let's see what we can do for ya. :)

LauraRozier:
I just updated it and added a nifty feature, drag&drop creation of new map objects.
Maybe this can help you add more smootly? :)

~~If this is a successful feature, I'll make it into a proper PR. :)~~
PR created.

Alexandr57:
PR is closed because the branch is not working. New PR already with all the corrections - https://github.com/Kromster80/kam_remake/pull/430

PR #324, opened 2017-03-01
New dynamic scripts functions: Format

reyandme:
Format - pascal's Format wrapper

_TimeGet - our KM_Utils.TImeGet wrapper to get time in milliseconds for some measurement purposes_

TimeGet is removed from this PR

### DIFF
--- a/src/scripting/KM_ScriptingUtils.pas
+++ b/src/scripting/KM_ScriptingUtils.pas
@@ -24,6 +24,8 @@ type
     function EnsureRangeI(aValue, aMin, aMax: Integer): Integer;
     function EnsureRangeS(aValue, aMin, aMax: Single): Single;
 
+    function Format(aFormatting: string; aData: array of const): string;
+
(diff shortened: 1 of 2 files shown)


reyandme:
TimeGet could be used for change game state which is unsafe for MP game if not used cautiously

src/scripting/KM_ScriptingUtils.pas
```diff
@@ -301,6 +303,20 @@ function TKMScriptUtils.EnsureRangeS(aValue, aMin, aMax: Single): Single;
 
 
 //* Version: 7000+
+//* Wrapper for pascal Format function
+//* Formats aFormatting string with specified aData array of parameters
+function TKMScriptUtils.Format(aFormatting: string; aData: array of const): string;
+begin
+  try
```
Kromster80: Let's add validation here. Compare number of passed parameters and number of % occurring in a string.

Issue #321, opened 2017-02-27
Changes to Actions.PlayerShareFog breaks backwards script compatibility


lewinjh:
@andreus791 We want to maintain backwards compatibility if possible so old r6720 scripts still work in the next release without script authors having to fix them all. Can we add a new Actions.PlayerShareFogCompliment? Or simply remove the compliment parameter since it can be achieved by running:
Actions.PlayerShareFog(A, B, True)
Actions.PlayerShareFog(B, A, True)
?

andreus791:
Yeah I forgot about back compatibility... Maybe new script with additional parameter will be better (like GiveField/GiveFieldAged).

reyandme:
fixed in #325

PR #325, opened 2017-03-02
Changes to Actions.PlayerShareFog breaks backwards script compatibility

andreus791:
Reverted old Actions.PlayerShareFog and added ShareFogCompliment to maintain backwards compatibility


reyandme:
This PR fixes #321

PR #327, opened 2017-03-02
Dynamic scripts. New MapTilesArraySet function

reyandme:
Updates array of tiles at once, which is much faster, then do update tile by tile with 3 functions MapTileSet/MapTileObjectSet/MapTileHeightSet, because some pathfinding recompute is done after every of this function. 

Specially for Toxic ^)

### DIFF
--- a/src/KM_Points.pas
+++ b/src/KM_Points.pas
@@ -62,6 +62,8 @@ type
   function KMInRect(aPoint: TKMPointF; aRect: TKMRectF): Boolean; overload;
   function KMRectArea(aRect: TKMRect): Integer;
   function KMRectMove(aRect: TKMRect; X,Y: Integer): TKMRect;
+  procedure KMRectAddPnt(var aRect: TKMRect; X,Y: Integer); overload;
+  procedure KMRectAddPnt(var aRect: TKMRect; aPoint: TKMPoint); overload;
 
   function KMGetDirection(X,Y: Integer): TKMDirection; overload;
   function KMGetDirection(X,Y: Single): TKMDirection; overload;
@@ -352,6 +354,22 @@ begin
   Result.Right  := aRect.Right + X;
   Result.Top    := aRect.Top + Y;
   Result.Bottom := aRect.Bottom + Y;
+end;
+
+
+procedure KMRectAddPnt(var aRect: TKMRect; X,Y: Integer);
+begin
+  KMRectAddPnt(aRect, KMPoint(X,Y));
+end;
+
+
+procedure KMRectAddPnt(var aRect: TKMRect; aPoint: TKMPoint);
+begin
+  if KMInRect(aPoint, aRect) then Exit;
+  aRect.Left    := Min(aPoint.X, aRect.Left);
+  aRect.Right   := Max(aPoint.X, aRect.Right);
+  aRect.Top     := Min(aPoint.Y, aRect.Top);
+  aRect.Bottom  := Max(aPoint.Y, aRect.Bottom);
 end;
 
 
--- a/src/KM_Points.pas
+++ b/src/KM_Points.pas
@@ -52,6 +52,7 @@ type
   function KMRectF(aPoint: TKMPointF): TKMRectF; overload;
   function KMRectF(aLeft, aTop, aRight, aBottom: SmallInt): TKMRectF; overload;
   function KMRectRound(aRect: TKMRectF): TKMRect;
+  function KMSameRect(aRect1, aRect2: TKMRect): Boolean;
   function KMRectGrow(aRect: TKMRect; aInset: Integer): TKMRect;
   function KMRectGrowTopLeft(aRect: TKMRect): TKMRect;
   function KMRectShinkTopLeft(aRect: TKMRect): TKMRect;
@@ -62,8 +63,8 @@ type
   function KMInRect(aPoint: TKMPointF; aRect: TKMRectF): Boolean; overload;
   function KMRectArea(aRect: TKMRect): Integer;
   function KMRectMove(aRect: TKMRect; X,Y: Integer): TKMRect;
-  procedure KMRectAddPnt(var aRect: TKMRect; X,Y: Integer); overload;
-  procedure KMRectAddPnt(var aRect: TKMRect; aPoint: TKMPoint); overload;
+  procedure KMRectIncludePoint(var aRect: TKMRect; X,Y: Integer); overload;
+  procedure KMRectIncludePoint(var aRect: TKMRect; aPoint: TKMPoint); overload;
 
   function KMGetDirection(X,Y: Integer): TKMDirection; overload;
   function KMGetDirection(X,Y: Single): TKMDirection; overload;
@@ -116,6 +117,8 @@ type
 const
   ZERO_POINT: TKMPoint = (X: 0; Y: 0);
   INVALID_MAP_POINT: TKMPoint = (X: -1; Y: -1);
+
+  KMRECT_INVALID_TILES: TKMRect = (Left: -1; Top: -1; Right: -1; Bottom: -1);
 
 
 implementation
@@ -279,6 +282,15 @@ begin
 end;
 
 
+function KMSameRect(aRect1, aRect2: TKMRect): Boolean;
+begin
+  Result := (aRect1.Left = aRect2.Left)
+        and (aRect1.Top = aRect2.Top)
+        and (aRect1.Right = aRect2.Right)
+        and (aRect1.Bottom = aRect2.Bottom);
+end;
+
+
 function KMRectGrow(aRect: TKMRect; aInset: Integer): TKMRect;
 begin
   Result.Left   := Math.Max(aRect.Left   - aInset, 0);
@@ -357,13 +369,13 @@ begin
 end;
 
 
-procedure KMRectAddPnt(var aRect: TKMRect; X,Y: Integer);
+procedure KMRectIncludePoint(var aRect: TKMRect; X,Y: Integer);
 begin
-  KMRectAddPnt(aRect, KMPoint(X,Y));
+  KMRectIncludePoint(aRect, KMPoint(X,Y));
 end;
 
 
-procedure KMRectAddPnt(var aRect: TKMRect; aPoint: TKMPoint);
+procedure KMRectIncludePoint(var aRect: TKMRect; aPoint: TKMPoint);
 begin
   if KMInRect(aPoint, aRect) then Exit;
   aRect.Left    := Min(aPoint.X, aRect.Left);
--- a/src/scripting/KM_ScriptingActions.pas
+++ b/src/scripting/KM_ScriptingActions.pas
@@ -4,7 +4,8 @@ interface
 uses
   Classes, Math, SysUtils, StrUtils, uPSRuntime,
   KM_CommonTypes, KM_Defaults, KM_Points, KM_Houses, KM_ScriptingIdCache, KM_Units,
-  KM_UnitGroups, KM_ResHouses, KM_HouseCollection, KM_ResWares, KM_ScriptingEvents;
+  KM_UnitGroups, KM_ResHouses, KM_HouseCollection, KM_ResWares, KM_ScriptingEvents,
+  KM_Terrain;
 
 
 type
@@ -89,6 +90,7 @@ type
     procedure Log(aText: AnsiString);
 
     function MapTileSet(X, Y, aType, aRotation: Integer): Boolean;
+    function MapTilesArraySet(aTiles: TKMTerrainTileBrief2Array; aRevertOnFail, aShowDetailedErrors: Boolean): Boolean;
     function MapTileHeightSet(X, Y, Height: Integer): Boolean;
     function MapTileObjectSet(X, Y, Obj: Integer): Boolean;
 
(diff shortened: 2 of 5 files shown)


LauraRozier:
Maybe squashing some commits might be a good idea?
Some tool like GitEye is nice for that task. (It's made using Eclipse, I know. But it does do it's job quite nicely.)

reyandme:
@thibmo Sorry, I made some initial commits for personal use. 
I use GitKraken. Its nice tool, I mostly satisfied with it. But squashing is working fine there when used for local repo. I believe its not "tool dependant", so its probably same with GitEye. If I am wrong, please correct me. 
When I push it onto remote branch its too late unfortunaly for squashing, some mess is going on after it. 

Also some fixes I prefer to make in different commits to be able to distinguish what was fixed in every of them (last 3 commits).

LauraRozier:
What I do for 'remote' squashing is a forced push. ( With TortoiseGit you can only do this using the 'Git Synch' tool AFAIK )
Not really a nice thing to do, but GH handles it perfectly fine.

Squashing is indeed quite similar between tools, example of the git CLI tool :
[git - Interactive rebase](https://blog.ona.io/general/2016/02/02/squashing-with-git-interactive-rebase.html)

I checked GitKraken before and couldn't find any squash function...
Maybe I didn't look properly. :)

Still looking for a Windows && GNU-linux tool that allows me to revise the squash ( Leaving some bits out for example ) and allow me to only squash specific commits, even if there is a commit in-between ( Which I might want to keep for some reason ) eg: squash commit 100, 101, 125 and 130 into a single commit.

Suggestions are welcome. :P

src/terrain/KM_Terrain.pas
```diff
@@ -585,31 +624,192 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+  DiagonalChangedTotal := False;
+  Rect := KMRect(-1, -1, -1, -1);
+  // Use separate HeightRect, because UpdateLight invoked only when Height is changed
+  HeightRect := KMRect(-1, -1, -1, -1);
+  ErrCnt := 0;
+
+  // make backup copy of Land only if we may need revert changes
+  if aRevertOnFail then
+  begin
+    SetLength(BackupLand, fMapY);
```
Kromster80: You can set it at once like so: SetLength(BackupLand, fMapY, fMapX);

src/terrain/KM_Terrain.pas
```diff
@@ -16,6 +16,27 @@ interface
...
+  TKMTileChangeType = (tctTerrain, tctHeight, tctObject);
+
+  TKMTileChangeTypeSet = set of TKMTileChangeType;
+
+  TKMTerrainTileBrief = packed record
+    Terrain: Byte;
+    Rotation: Byte;
+    Height: Byte;
+    Obj: Byte;
+    Contains: TKMTileChangeTypeSet;
```
Kromster80: Better to rename to ChangeSet: TKMTileChangeTypeSet;

Kromster80 (changes requested):

reyandme:
`git - Interactive rebase`

Thank you for link @thibmo. I agree, its much better to squash all these commits into one.

reyandme:
@thibmo Well, squashing works well in GitKraken after Force Push.
But its possible to squash only to the last remote commit. So you cant squash first before last and second before last into one commit, you have to join them to the last one only

reyandme:
@thibmo btw, as we talked here: can you please fix our dedicated server? Because after your PR #196 its not compile anymore.
There are 2 projects AFAIR - GUI version and CLI one.
Thanks :)

src/terrain/KM_Terrain.pas
```diff
@@ -16,6 +16,27 @@ interface
                     vu_NESW);   //Vertex is used NE-SW like this: /
   TFenceType = (fncNone, fncCorn, fncWine, fncHousePlan, fncHouseFence);
 
+  TKMTileChangeType = (tctTerrain, tctHeight, tctObject);
+
+  TKMTileChangeTypeSet = set of TKMTileChangeType;
+
+  TKMTerrainTileBrief = packed record
```
lewinjh: It's less efficient to use packed records unless we care about padding for saving/loading (or RAM usage in really big arrays). Packing forces the CPU to do more work extracting each element since they are not nicely aligned. None of these records need packing as far as I can see.

src/terrain/KM_Terrain.pas
```diff
@@ -215,6 +242,8 @@   TKMTerrain = class
     procedure UpdatePassability(aRect: TKMRect); overload;
     procedure UpdatePassability(Loc: TKMPoint); overload;
 
+    procedure UpdateWalkConnect(const aSet: array of TWalkConnect; aRect: TKMRect; aDiagObjectsEffected: Boolean);
```
lewinjh: Why did you make this public? It's an internal terrain thing so lets leave it private.

src/terrain/KM_Terrain.pas
```diff
@@ -585,31 +624,189 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+      X := I + 1;
+      Y := J + 1;
+
+      HasErrorOnTile := False;
+      ErrorTypesOnTile := [];
+
+      // Update terrain and rotation if needed
+      if tctTerrain in T.ChangeSet then
+      begin
+        if TileInMapCoords(X, Y) and InRange(T.Terrain, 0, 255) and InRange(T.Rotation, 0, 3) then
```
lewinjh: Rather than checking TileInMapCoordinates every time, lets just check that High(aTiles) <= fMapY and High(aTiles[I]) <= fMapX before each loop (and only loop to the minimum of these two).

InRange(T.Terrain, 0, 255) is a tautology when your data type is byte ;)

src/scripting/KM_ScriptingActions.pas
```diff
@@ -1981,6 +1983,73 @@ function TKMScriptActions.MapTileSet(X, Y, aType, aRotation: Integer): Boolean;
...
+//*    Height: Byte;   // Heigth (0..100)
+//*    Obj: Byte;      // Object (0..255)
+//*    ChangeSet: TKMTileChangeTypeSet; // Set of changes. F.e. if we want to change terrain type and height, than ChangeSet should contain tctTerrain and tctHeight
+//*  end
+//* where TKMTileChangeTypeSet = set of TKMTileChangeType
+//* where TKMTileChangeType = (tctTerrain, tctHeight, tctObject) // Determines what should be changed on tile
+//* Note: aTiles elements should start from 0, as for dynamic array. So f.e. to change map tile 1,1 we should set aTiles[0][0].
+//* Note: Errors are shown as map tiles (f.e. for error while applying aTiles[0][0] tile there will be a message with for map tile 1,1)
+
+function TKMScriptActions.MapTilesArraySet(aTiles: TKMTerrainTileBrief2Array; aRevertOnFail, aShowDetailedErrors: Boolean): Boolean;
```
lewinjh: Maybe it would be better to allow the user to specify a starting X and Y? That way you can update a small area in the centre of the map without allocating a massive array. So the user is effectively providing a rectangle of {StartX, StartY, High(aTiles[I]), High(aTiles)}.

lewinjh (changes requested):
I'm wondering if it's necessary to allow the user to specify TKMTileChangeTypeSet for every single tile rather than once for the entire action? It means more work for the scripter to initialise the array and it makes the array larger.

Can you think of cases where you would want to change different sets of properties on each tile in a single call?

src/scripting/KM_ScriptingActions.pas
Kromster80: I'm also advocating for adding of Width/Height and changing to 1D array (aka simple pointer at 0th element), so we don't have to deal with possible non-rectangular 2D arrays

reyandme:
> Can you think of cases where you would want to change different sets of properties on each tile in a single call?

Yes, its possible situation. f.e. Toxic in his script currently update some tiles with:
- only terrain + rotation
- terrain + rotation + height (all mountains, somewhere on terrain)
- terrain + rotation + object (where object is placed)
- terrain + rotation +height + object (same as all above)

Other solution - add separate functions to update bunch of tiles for terrain+rotation/objects/height. But I think its ok, if scripters will do a bit more work. They can easily set all the parameters in some function and make it more or less convinient.

We can assume, that terrain and rotation is set by default. But I personally prefer to let scripter decide what to add and what not to add. (without "default" value for TKMTileChangeTypeSet). He can easily do that in some function (or functions)

src/terrain/KM_Terrain.pas
reyandme: yep, sorry, it was occasional refactoring error.

src/terrain/KM_Terrain.pas
reyandme: @Kromster80 suggested to me to use packed record, because I had some problems with memory - we also make a backup copy of all Land array and after it we invoke UpdateWalkConnect on all of the array (possible 255*255 Map). So that is why.
Probably we can remove packed now, sinse some optimisations was added afterwards.
@Kromster80 what do you think ?

src/scripting/KM_ScriptingActions.pas
reyandme: I like 1D array idea. Actually it was initially implemented as 1D.

src/terrain/KM_Terrain.pas
reyandme: ok

reyandme:
After this PR merge I'll be able to start work on Toxic's RMG generator boiler-plate. Because some of the code from this PR could be used in it or interfere with it.

src/KM_Points.pas
```diff
@@ -62,6 +62,8 @@   TKMTriMesh = record
   function KMInRect(aPoint: TKMPointF; aRect: TKMRectF): Boolean; overload;
   function KMRectArea(aRect: TKMRect): Integer;
   function KMRectMove(aRect: TKMRect; X,Y: Integer): TKMRect;
+  procedure KMRectAddPnt(var aRect: TKMRect; X,Y: Integer); overload;
```
Kromster80: Should be called "KMRectIncludePoint", since we are including this point into the Rect.

src/terrain/KM_Terrain.pas
```diff
@@ -55,6 +75,7 @@   TKMTerrainTile = record
   {Class to store all terrain data, aswell terrain routines}
   TKMTerrain = class
   private
+    fCnt: Cardinal;
```
Kromster80: Unclear variable name

src/terrain/KM_Terrain.pas
```diff
@@ -585,31 +622,210 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+
+// Try to set an array of Tiles from script. Set terrain, rotation, height and object.
+// Update Passability, WalkConnect and Lighting only once at the end.
+// This is much faster, then set tile by tile with updates on every change
+//
+// Returns True if succeeded
+// use var for aTiles. aTiles can be huge so we do want to make its local copy. Saves a lot of memory
+function TKMTerrain.ScriptTrySetTilesArray(var aTiles: array of TKMTerrainTileBrief; aRevertOnFail: Boolean; var aErrors: TKMTerrainTileChangeErrorArray): Boolean;
+
+  function IsRectInit(aRect: TKMRect): Boolean;
```
Kromster80: Should be in KM_Points.KMRectIsInit ?

src/terrain/KM_Terrain.pas
```diff
@@ -585,31 +622,210 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+  begin
+    Result := (aRect.Left <> -1) and (aRect.Right <> -1)
+          and (aRect.Top <> -1) and (aRect.Bottom <> -1);
+  end;
+
+  procedure UpdateRect(var aRect: TKMRect; X, Y: Integer);
```
Kromster80: Should be in KM_Points.KMRectIncludePoint ?

src/terrain/KM_Terrain.pas
```diff
@@ -585,31 +622,210 @@ function TKMTerrain.ScriptTryObjectSet(X, Y: Integer; aObject: Byte): Boolean;
...
+end;
+
+
+// Try to set an object from the script. Failure is an option
+function TKMTerrain.ScriptTrySetTileObject(X, Y: Integer; aObject: Byte): Boolean;
+begin
+  Result := False;
+  if not TileInMapCoords(X, Y) then
+    Exit;
+  Result := TrySetTileObject(X, Y, aObject);
```
Kromster80: One line could be better instead:
Result := TileInMapCoords(X, Y) and TrySetTileObject(X, Y, aObject);

src/terrain/KM_Terrain.pas
reyandme: I added KMSameRect and KMRECT_INVALID_TILES, same as for Points. Then its better to call KMSameRect(Rect, KMRECT_INVALID_TILES)

src/terrain/KM_Terrain.pas
reyandme: I do like this kind of 'hidden' logic. KMRectIncludePoint should be very simple - we should be able to include point to the rect even for (-1,-1,-1,-1) rect - its invalid for map tiles, but not invalid in general (same as for Points)

Kromster80:
I guess it's okay.
@lewinjh what do you say?

PR #328, opened 2017-03-02
Update position in sp game list on maped game save

reyandme:
Could be usufull for mapmakers, when they edit SP map (or campaign map, which is usually edited as SP map). Then if they want to test their changes map will be already selected in list.

For all other players it could not add any difficulties, because it not often to play same SP map all the time and also they probably do not use MapEd so much.

If mapmakers want to check their MP map changes, they could use 'favorite maps' feature - they are updated on maped resave


Kromster80:
It would be much better to add a button into MapEd [Play this mission] instead.

reyandme:
Yes, it's a good alternative. We can add this button to the last menu (save/load)

Kromster80:
Let's do that instead :)

reyandme:
Ok !

PR #330, opened 2017-03-04
Maped new tool: Universal eraser

reyandme:
Eraser can delete units, groups, houses, roads, fields, winefields and terrain objects. 

Units are selected by unit "selecting rect", while all other - by cell under the cursor. 

When there are several thing to delete (f.e. road and terrain object and unit) then next order is applied, first deletes on first click, second on second etc: unit -> object -> road. 
If also Shift was pressed while deleting then it deletes all, what is possible (unit and terrain object and road, f.e.)

Deleting object is highlighted (with red or qual cell).

Here it is a small video, shows how does eraser work: [video](https://puu.sh/utEl2/d4d594ac37.mp4)

I used red cross as an icon. It would be nice to have special cursor for that eraser, same as we have for "paint bucket" feature. Malin draw one (and similar icon also), but sure if it looks good: 
cursor

Also I could not stop myself from moving cmPaintBucket methods from TKMMapEdInterface to TKMMapEditor, where they should be. For paint bucket same logic with "painting" all under the cursor if Shift was pressed.


Alexandr57:
It looks good but did not quite understand. If we want to erase the Unit and he was on the road. chtol Road also erased?

reyandme:
@Alexandr57 if you hold Shift - then yes, everything on tile will be deleted. If you do not press shift - then only one thing will be deleted in next order: unit -> object -> road. I think its better you just try it

reyandme:
Is there any chance this PR will be reviewed one day ?

src/KM_MapEditor.pas
```diff
@@ -215,12 +218,14 @@ function TKMMapEditor.HumanCount: Integer;
 end;
 
 
-procedure TKMMapEditor.UpdateField(aInc: Integer; aCheckPrevCell: Boolean);
+//aStageIncrement - stage increment, can be negative
```
Kromster80, line 221: This is a bad sort of comment. It says exactly what argument name says - that this is a "stage" and "increment" and it has signed type.
Could you make it more "helpful" instead?

src/KM_MapEditor.pas
```diff
@@ -249,6 +254,97 @@ procedure TKMMapEditor.UpdateField(aInc: Integer; aCheckPrevCell: Boolean);
 end;
 
 
+procedure TKMMapEditor.EraseObject(aEraseAll: Boolean);
+var Obj: TObject;
+    P: TKMPoint;
+    IsDeleted: Boolean;
+begin
+  IsDeleted := False;
```
Kromster80: Let's make it more uniform and with less variables?

For each next object:

if %object_under_cursor% then
begin
  %delete_object%
  if not aEraseAll tehn Exit;
end;

if %next_object_under_cursor% then
 ....

and so on.

src/KM_MapEditor.pas
```diff
@@ -249,6 +254,97 @@ procedure TKMMapEditor.UpdateField(aInc: Integer; aCheckPrevCell: Boolean);
...
+end;
+
+
+//Change owner for specified object
+//returns True if owner was changed successfully
+function TKMMapEditor.ChangeObjectOwner(aObject: TObject; aOwner: TKMHandIndex): Boolean;
+var House: TKMHouse;
+begin
+  Result := False;
+  if (aObject <> nil) then
```
Kromster80: Better to write "if Obj = nil then Exit;"

Kromster80 (changes requested):
Reviewed 1st half

src/KM_MapEditor.pas
reyandme: I am not sure what else can we say about this method argument. It is really looks pretty obvious why it is needed, so that is why there was no any comment about it before.
The reason why I added this comment is that I want to emphasise the fact it could be negative (so it will actually work as decrement, not as increment, which was not obvious from its name).
If you have any suggestions about this comment, I can update it with ease :)

src/KM_MapEditor.pas
reyandme: sure

PR #331, opened 2017-03-04
Scripting events

andreus791:


reyandme:
Mb we can also add OnGroupOrderFeed / OnGroupOrderWalk / OnGroupOrderStorm ? 
Not sure all of them will be used, but you never know. ScriptMakers can ask then, why we introduced some order functions and did not introduce others.

PR #332, opened 2017-03-04
Unit group: feed or split command should not cancel group previous order

reyandme:
If group has some order (walk/attack unit/house) then it is usually continue it after some group command was initiated (change rows count, turn left or right, link). But for some reason group get halt order on feed and split command. Its very unpleasant - player send his troops to some point, then he wanted to feed them or split - and group stop going to that point, making formation instead.

This PR fixes this issue.

### DIFF
--- a/src/units/KM_UnitGroups.pas
+++ b/src/units/KM_UnitGroups.pas
@@ -1241,8 +1241,8 @@ begin
   case fOrder of
     goNone:         OrderHalt(False);
     goWalkTo:       OrderWalk(fOrderLoc.Loc, False);
-    goAttackHouse:  if OrderTargetHouse <> nil then OrderAttackHouse(aGroup.OrderTargetHouse, False);
-    goAttackUnit:   if OrderTargetUnit <> nil then OrderAttackUnit(aGroup.OrderTargetUnit, False);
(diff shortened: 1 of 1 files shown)


src/units/KM_UnitGroups.pas
```diff
@@ -1232,19 +1231,37 @@ procedure TKMUnitGroup.OrderNone;
 end;
 
 
-//Repeat last order e.g. if new members have joined
-procedure TKMUnitGroup.OrderRepeat;
+//Repeat order after specified aGroup or repeat self order if no aGroup is specified
+procedure TKMUnitGroup.OrderRepeat(aGroup: TKMUnitGroup);
```
lewinjh: Lets leave the old OrderRepeat as is and add a new procedure, CopyOrderFrom(aGroup). Less confusing that way and it's a short procedure so not much duplication.

lewinjh (changes requested):
I guess we did it that way because that's what KaM did. I agree with you, it's better not to halt in these situations.

src/units/KM_UnitGroups.pas
reyandme: ok, I agree

lewinjh (approved):

PR #340, opened 2017-03-06
Allow to set middle mouse as any shortcut (same as any other key). Drag scrolling refactoring

reyandme:
Middle mouse is used for map drag scrolling. I believe its not very popular feature, so probably some player would like to change it to something else.

This PR provide such oppotunity. For drag scrolling also any hotkey is allowed now, not only middle mouse btn.

Some refactoring due to drag scrolling has been done.


KM_KeyIDs.inc
```diff
@@ -3,92 +3,94 @@ SC_SCROLL_RIGHT = 1; //Scroll Right
 SC_SCROLL_UP = 2; //Scroll Up
 SC_SCROLL_DOWN = 3; //Scroll Down
 
-SC_ZOOM_IN = 4; //Zoom In
-SC_ZOOM_OUT = 5; //Zoom Out
-SC_ZOOM_RESET = 6; //Reset Zoom
+SC_MAP_DRAG_DCROLL = 4; //Map drag scroll
```
Kromster80: Typo?

src/gui/KM_InterfaceDefaults.pas
```diff
@@ -52,12 +52,12 @@   TKMUserInterfaceCommon = class
     property MyControls: TKMMasterControl read fMyControls;
     procedure ExportPages(aPath: string); virtual; abstract;
 
-    procedure KeyDown(Key: Word; Shift: TShiftState); virtual; abstract;
+    function KeyDown(Key: Word; Shift: TShiftState): Boolean; virtual; abstract;
```
Kromster80: How about if we change this and alikes into "procedure KeyDown(Key: Word; Shift: TShiftState; var aHandled: Boolean); virtual; abstract;" ?

src/gui/KM_InterfaceGamePlay.pas
```diff
@@ -3810,6 +3777,14 @@ procedure TKMGamePlayInterface.UpdateStateIdle(aFrameTime: Cardinal);
 end;
 
 
+function TKMGamePlayInterface.IsDragScrollingAllowed: Boolean;
+begin
+  Result := not (gGame.IsPaused and (fUIMode in [umSP, umMP]))
```
Kromster80: When overriding a methods, please always include "inherited". Even in this trivial case, when it is always True in parent class.

src/gui/KM_InterfaceGame.pas
```diff
@@ -149,6 +158,12 @@ constructor TKMUserInterfaceGame.Create(aRender: TRender);
 
   fMinimap := TKMMinimap.Create(False, False);
   fViewport := TKMViewport.Create(aRender.ScreenX, aRender.ScreenY);
+
+  fDragScrolling := False;
+  fDragScrollingCursorPos.X := 0;
+  fDragScrollingCursorPos.Y := 0;
+  fDragScrollingViewportPos := ZERO_POINT_F;
```
Kromster80: You know that in constructors all fields are initialized with "zeroes" right?

Adding a ZERO_POINT_F constant just for this is a bit unjustified.

src/KM_Points.pas
```diff
@@ -113,6 +113,7 @@   TKMTriMesh = record
 
 const
   ZERO_POINT: TKMPoint = (X: 0; Y: 0);
+  ZERO_POINT_F: TKMPointF = (X: 0.0; Y: 0.0);
```
Kromster80: See comment below.
On a sidenote: let's refactor these into KMPOINT_ZERO and KMPOINT_INVALID, so they share the same starting letters and be logically grouped.

Kromster80 (changes requested):

KM_KeyIDs.inc
reyandme: yes

src/gui/KM_InterfaceDefaults.pas
reyandme: ok, both options are fine for me

src/gui/KM_InterfaceGamePlay.pas
reyandme: yeah, sure

src/gui/KM_InterfaceGame.pas
reyandme: I know, but anyway I prefer to explicitly set them. So when looking through the code (with Ctrl+Shift+Enter f.e.) I can see its initialization value.

Here I just replaced 

`fDragScrollingViewportPos.X := 0;
fDragScrollingViewportPos.Y := 0;`
with 
`fDragScrollingViewportPos := ZERO_POINT_F;`

ALso I was planning to do small refactoring: replace all KMPoint(0, 0) and KMPointF(0.0, 0.0) with new constants. Don't you mind it ?

src/KM_Points.pas
reyandme: Ok, I can rename them. But better in refactoring PR, which I mentioned above

reyandme: KMPOINT_INVALID is a bit confusing, because point (-1, -1) is not invalid (we can use it as some adjustment f.e.). But its invalid point on Map. That is why I initially name it INVALID_MAP_POINT.

So may be KMPOINT_INVALID_ON_MAP or smth ?

src/gui/KM_InterfaceGame.pas
Kromster80: Yes, that would be good :)

src/KM_Points.pas
Kromster80: If it fits, maybe KMPOINT_INVALID_TILE

reyandme: ok

reyandme:
@Kromster80 I've done some refactoring.
Points refactoring I will do in separate PR, as I mentioned earlier.

reyandme:
@Kromster80 refactoring is ok ? 
I want to use this PR for objects palette in MapEd. This was initial reason of adding middle mouse custom hotkeys.

reyandme:
I've updated this PR to actual state.
@Kromster80 can you check and merge it then?

PR #343, opened 2017-03-07
Dynamic scripts: update to current version of remObjects PascalScript. Added ability to include scripts. Added pascal style directives. Show exact error position in script.

reyandme:
PascalScript was updated to current master branch of remObjects [PascalScript](https://github.com/remobjects/pascalscript).
I checked commits from ~may 2014 - mostly there were some refactorings and bugfixes, no any significant change was made. But anyway its worth updating I believe, especially because I added 2 missing in our repository files: uPSPreProcessor.pas and uPSDebugger.pas.

PreProcessor is used to add ability to include script files one onto another with directives: 
{$I filename.script} or {$INCLUDE filename.script} 
$I and $INCLUDE are synonims. 

Also PreProcessor add other directives functionality: 
${DEFINE MY_VAR} 
${UNDEF MY_VAR}
${IFDEF MY_VAR}
${IFNDEF MY_VAR}
${ELSE}
${ENDIF}
They work similar to well known pascal directives.

Also PreProcessor opens ability to create our own directives. 
Smth like ${DEBUG ON}, {$SHOW_ERROR_LINE}, ${COPY_ERROR_LINE_CODE_TO_CLIPBOARD}, ${HIDE_SCRIPT_ERRORS} etc, or anything else, that could be helpful for scriptmakers. We just need to add special handler method in out scripting class.
Any suggestions about new directives ?

uPSDebugger.pas is used to show exact script code position, where error occurs. 
With our new includes system I added also info about file, where error occurs and row/col of error in that file.


Kromster80:
I have questions, since there are more than several logical commits in this PR:

 - What is the reasoning behind adding "{$WARN IMPLICIT_STRING_CAST OFF}" ?
 - What is the reasoning behind changing fScripting into gScripting?

Please add more comments along the code, explaining "What and Why", e.g. new code in "procedure TKMScriptEvents.DoProc(.." or why there was a need in wrapping this into a string "String(IntToStr(aValues[I]))" ?

reyandme:
> What is the reasoning behind adding "{$WARN IMPLICIT_STRING_CAST OFF}" ?

There are about 10 warnings W1057 in this Scripting unit, mostly because we interact with pascal script classes (uPSPreProcessor/uPSDebugger). So there was an alternative - turn this warning off or add everywhere implicit cast. I choose first. I think there is no a big problem in it.

> What is the reasoning behind changing fScripting into gScripting

Because of the changes in PascalScript we have to use regular procedure instead of method pointer in their method handlers (Compiler.OnUses/OnUseVariable/OnExportCheck; PreProcessor.OnNeedFile), so I added fScripting method wrappers for that. But to access fScripting from there I switched to gScripting. 
Probably we could simply archive the same by invoking gGame.Scripting from there. I can fix it

> new code in "procedure TKMScriptEvents.DoProc(.."

This new code gets error position from debugger. I will add comment 

> why there was a need in wrapping this into a string "String(IntToStr(aValues[I]))

String() could be removed (I forget to remote it) with added "{$WARN IMPLICIT_STRING_CAST OFF}".  But without it we will get 2 warning in this unit. Good example of what we will have if remove this $WARN directive, btw. 
Warnings help me a lot to fix small issues after compilation, but not these kind of warnings. Its just about, that we have to cast AnsiString to String explicitly, which I believe is not a big deal here.

Kromster80:
You can disable string-cast warnings in your IDE too.

src/scripting/KM_Scripting.pas
```diff
@@ -200,6 +277,41 @@ procedure TKMScripting.LoadFromFile(aFileName: UnicodeString; aCampaignDataTypeF
...
+                                  [ExtractFileName(aFileName), ExtractFileName(aCallingFileName)]));
+
+  S := ExtractFilePath(aCallingFileName);
+  if S = '' then S := ExtractFilePath(ParamStr(0));
+  aFileName := AnsiString(S) + aFileName;
+  if FileExists(aFileName) then
+  begin
+    aOutput := ReadTextA(aFileName);
+
+    IncludeInfo := TKMScriptFileIncludeInfo.Create;
```
Kromster80: Is there a corresponding "Free" call?

Kromster80:
As discussed in Skype - needs refactoring and bugfixing

lewinjh:
{$INCLUDE filename.script} will cause problems since extra script files will not be handled correctly by the game (not included in multiplayer transfers, CRCs, crash reports, moving/renaming maps, etc.). So we either need to not allow this or find a way to contain all of these extra files (like WAV files they would need to follow the pattern of <MapName>.name.script) We would also want to block including from folders outside of the  current map's folder.

I'm not sure DEFINE preprocessor directives are necessary but I guess they don't hurt.

Showing the exact error position is good.

reyandme:
@lewinjh 

> {$INCLUDE filename.script} will cause problems since extra script files will not be handled correctly by the game (not included in multiplayer transfers, CRCs, crash reports, moving/renaming maps, etc.).

I think its worth adding support of includes directives, because there some really big scripts. Also some script "libraries" could appear -  f.e. some collections of usefull function, which we did not add to 'Utils' scripting unit for some reason.

> we either need to not allow this or find a way to contain all of these extra files (like WAV files they would need to follow the pattern of .name.script)

Its a possible way, but do we need to restrict users with include names ? As for "libraries" - they better have same name over different scripts.

> We would also want to block including from folders outside of the current map's folder.

Yes, I agree, I'll fix it.

> I'm not sure DEFINE preprocessor directives are necessary but I guess they don't hurt.

I asked @andreus791 about it (I was also very sceptical about directives usage). He said that they could be usefull in his scripts.
Also I still thinking that we can add some new directives for something. Any ideas are appreciated. 
@andreus791 do you have some ?

lewinjh:
I agree libraries would be good.

In the past we tried to keep all files related to a map named in the same fashion starting with the map name, that's why with WAVs we insist on MapName.CustomName.wav. I suppose we could relax this restriction (I can't remember if there was an important reason for it). Although we need to make sure all such files are included in MP transfers, CRCs, crash reports, etc.

reyandme:
While refactoring one small problem occurs: to get all files to transfer, or to get map CRC - we have to preprocess script file. But scripting is currently placed in TKMGame, so there is no scripting at the moment, when its already needed.
I suggest to move fScripting out of TKMGame to TKMGameApp. In game we can reuse its instance, since there is always only one gGame at the moment. Also using global gScripting would be usefull.
@Kromster80 @lewinjh is it ok ?

Kromster80:
What kind of pre-processing is needed? Can it be done in a "class function" of scripting?

The idea behind modules is this:
 - gMain is man application level, all generic methods are there
 - gGameApp is a game in broad sense. It could be RTS, or Shooter, or Racing.
 - gGame is actual instance of our RTS game.

We cut many corners for our convenience by using gTerrain and gHands and such, but truth be told, they all should be within gGame.  
With above in mind, scripting belongs to gGame level - it's instantiated once per game, works only in game, if we run 2 gGames in parallel, each will need it's own Scripting.

reyandme:
Ok, I see, thx for explanation.

Pre-processing can be made in separate class function without any problem (with small refactoring). Actually I would do that also, but previously you suggest me to avoid class functions (for some other task I wanted to add them IIRC). For me personally its okay to have them

Kromster80:
Class functions are okay for me too. I must have been suggesting against them in other place based on some other reasoning.

reyandme:
I've done lots of refactoring. 
Included files now in plain array structure (instead of tree structure) for simplicity reasons.
Instead of using class methods I added TKMScriptingPreProcessor class, and also TKMScriptingErrorHandler class. 
PreProcessor can be used from TKMScripting, or as separate instance - for calculating Map CRC and detecting attached files for net transfer. For crash reports and maps resave (from SP to MP, f.e.) I did not use PreProcessor, just attach/copy all script files, if main script file was detected. For crash reports - because they can be used in some units, so could be an error reason; for map resave - may be user wanted to include some of his scripts into main one later. 

Error handler can be used from PreProcessor or from Scripting, that is why it was needed to make it as a separate class.

In uPSPreProcessor unit I was forced to change TPSOnNeedFile and TPSOnProcessDirective types from procedure to method pointer. Before (in r6720) it was method pointer. In current master branch rPascalScript moved to procedure type for such event handlers - same for Compiler: OnUses/OnUseVariable/OnExportCheck. For compiler I used that fact, that we can not have 2 scripting instances (because we always have 1 gGame instance). But for PreProcessor it was akward, I could not find a way to initialize PreProcessor instance and let it use handler, which will able to access same instance. I believe there no such way by definition of procedure type, which is unrelated to any particular instance.

> all such files are included in MP transfers, CRCs, crash reports, etc.

Implemented

Error messages were refactored.

reyandme:
There is still small issue with showing compile errors/hint in the included files or after some of DEFINE directives - compile error row number shown script is all in one text (as it is after pre-processing).
Solution is to use same approach, as I did when show exact runtime error position in included file.

reyandme:
Is there any chance this PR will be reviewed one day ?

reyandme:
I found an error in my implementation, fixed in my repo in commit:
https://github.com/reyandme/kam_remake/commit/022255d15da3fd4832a6ddaae4d830bb4ba609aa
Did not put it into this thread, as It looks quite dead and not responding.

PR #338, opened 2017-03-06
Fixed return of color through PlayerColorText function in dynamic scripts

Alexandr57:
PlayerColorText function returns the wrong shade of color
More details on the screenshot:
image
1 Variant - Result := IntToHex(aColor and $00FFFFFF, 6);
2 Variant - Result := AnsiString(Format('%.6x', [FlagColorToTextColor(aColor) and $FFFFFF]));

PR #344, opened 2017-03-08
Revert "Fixed return of color through PlayerColorText function in dynamic scripts"

lewinjh:
Reverts Kromster80/kam_remake#338


Alexandr57:
@lewinjh 
What for?
Previous function is not the same color returns. It returns a lighter color

PR #345, opened 2017-03-08
Updating or adding a dynamic script analogue PlayerColorText

Alexandr57:
At this stage, I just updated the function of the TextbookText so that it returns not only the clarified color but also the original + so that the function remains compatible with the old scripts.
If you do not like it, I can add a separate one.
If you say that it's not obvious then In the wiki you can update

Here's what it looks like:

Using:
```
procedure OnMissionStart;
begin
	Actions.OverlayTextSet(0, '[$'+States.PlayerColorText(0)+']1 Variant|[$'+States.PlayerColorText(0, false)+']2 Variant');
end;
```

### DIFF
--- a/src/scripting/KM_ScriptingStates.pas
+++ b/src/scripting/KM_ScriptingStates.pas
@@ -1045,7 +1045,7 @@ begin
 end;
 
 
-//* Version: 4758
+//* Version: 4758 Update 7000+
 //* Get players color as text in hex format
 //* Result: Player color
 function TKMScriptStates.PlayerColorText(aPlayer: Byte;
@@ -1059,7 +1059,7 @@ begin
       if bLight then
         Result := AnsiString(Format('%.6x', [FlagColorToTextColor(gHands[aPlayer].FlagColor) and $FFFFFF]))
       else
-        Result := IntToHex(gHands[aPlayer].FlagColor and $FFFFFF, 6)
+        Result := AnsiString(IntToHex(gHands[aPlayer].FlagColor and $FFFFFF, 6))
     end
     else
     begin
--- a/src/scripting/KM_ScriptingStates.pas
+++ b/src/scripting/KM_ScriptingStates.pas
@@ -83,7 +83,7 @@ type
     function PeaceTime: Cardinal;
 
     function PlayerAllianceCheck(aPlayer1, aPlayer2: Byte): Boolean;
-    function PlayerColorText(aPlayer: Byte): AnsiString;
+    function PlayerColorText(aPlayer: Byte; bLight: Boolean = True): AnsiString;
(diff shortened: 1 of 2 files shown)


Alexandr57:
Requires the views of @lewinjh

src/scripting/KM_ScriptingStates.pas
```diff
@@ -1044,17 +1045,21 @@ function TKMScriptStates.StatResourceProducedMultipleTypesCount(aPlayer: Byte; a
...
 
-//* Version: 4758
+//* Version: 4758 Update 7000+
 //* Get players color as text in hex format
 //* Result: Player color
-function TKMScriptStates.PlayerColorText(aPlayer: Byte): AnsiString;
-begin
-  try
+function TKMScriptStates.PlayerColorText(aPlayer: Byte;
+  bLight: Boolean): AnsiString;
```
reyandme, line 1052: We use 'a' prefix for parameters, so pls change 'bLight' to 'aLight'.
Also why did you start new line here ?
(Мы используем префикс 'a', поэтому замени 'bLight' на 'aLight'.
Также почему ты добавил перенос строки ? Он здесь не нужен)

src/scripting/KM_ScriptingStates.pas
```diff
@@ -545,6 +545,7 @@ function TKMScriptStates.PlayerAllianceCheck(aPlayer1, aPlayer2: Byte): Boolean;
 end;
 
 
+
```
reyandme, line 548: No need for empty line
(Лишняя пустая строка)

reyandme (changes requested):

reyandme:
I added comments regarding code style.
Anyway @lewinjh should approve this changes first.

Alexandr57:
Not only @lewinjh  but @Kromster80
Kakto @Kromster80  is not enthusiastic but as already said it does not change the original function. So the old scripts, as seen on the screenshot, will work as well.
(Не только @lewinjh  но и @Kromster80 
Както @Kromster80  не в восторге но как уже сказано она не изменит оригинальную функцию. Так что старые скрипты как видно на скрине будут работать также.)

lewinjh:
Why do you need the original colour in the scripts? What can it be used for? As far as I'm aware the only use for colours in the script is to colour GUI text. If you use the original colours for text they are unreadable. This is the reason why we desaturate and lighten the colours for the GUI and script, since it makes unreadable colours like black readable, and highly saturated colours like red less hard on the eyes. The colours are still very close to the original and will not cause confusion.

For example compare black text with the original colour and the GUI/script version.

Alexandr57:
In some cases, the original color is very useful. Okay close but completely forbid using the original colors you can not. Personally, I hoped that there would be such a function. Well then, I will continue to use the color picker utility

PR #352, opened 2017-03-11
Refactoring: KMPoint(0,0) -> KMPOINT_ZERO

reyandme:
Also 
KMPointF(0,0) -> KMPOINT_ZERO_F
and rename INVALID_MAP_POINT -> KMPOINT_INVALID_TILE
As discussed in #340.

### DIFF
--- a/src/KM_Points.pas
+++ b/src/KM_Points.pas
@@ -112,8 +112,9 @@ type
 
 
 const
-  ZERO_POINT: TKMPoint = (X: 0; Y: 0);
-  INVALID_MAP_POINT: TKMPoint = (X: -1; Y: -1);
+  KMPOINT_ZERO: TKMPoint = (X: 0; Y: 0);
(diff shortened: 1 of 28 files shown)


src/KM_Points.pas
```diff
@@ -112,8 +112,9 @@   TKMTriMesh = record
 
 
 const
-  ZERO_POINT: TKMPoint = (X: 0; Y: 0);
-  INVALID_MAP_POINT: TKMPoint = (X: -1; Y: -1);
+  KMPOINT_ZERO: TKMPoint = (X: 0; Y: 0);
+  KMPOINT_ZERO_F: TKMPointF = (X: 0.0; Y: 0.0);
```
Kromster80: KMPOINTF_ZERO could be better with the naming scheme

reyandme: What do you mean with 'naming scheme'

PR #354, opened 2017-03-14
Showing the size of the map next to the name

Alexandr57:
Added displaying size map next to its name
image

### DIFF
--- a/src/gui/KM_InterfaceMapEditor.pas
+++ b/src/gui/KM_InterfaceMapEditor.pas
@@ -418,7 +418,7 @@ begin
 
   UpdatePlayerSelectButtons;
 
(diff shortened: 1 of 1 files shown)


Kromster80:
This is not kind of info that should be ALWAYS visible.
Besides, it is easy to check map size just by moving cursor over bottom-right map corner.
I'm in favor of rejecting this PR.

Alexandr57:
In fact, this is useful information in case you made a mistake in the size.
And that each time scroll down the map to find out the size.

Kromster80:
"In fact, this is useful information in case you made a mistake in the size.
And that each time scroll down the map to find out the size."

How often do you make "mistakes in the size", that you have to "each time scroll down the map to find out the size."

PR #357, opened 2017-03-16
MapEd: added objects palette

reyandme:
Added Objects palette for Map Editor, as suggested [here](http://www.knightsandmerchants.net/forum/viewtopic.php?f=24&t=3207&p=48027&hilit=palette#p48029)

Palette can be opened via button on terrain objects panel or via hotkey from anywhere - you can keep any other panel opened. On kam.net forum @Kromster80 suggested to allow open it only 'whenever you are placing terrain objects on a map', but I think its much more convinient for mapmaker to open it whenever he wants. 
Default hotkey is middle mouse button.

On palette you can see all the objects. If they do not fit into screen, then scroll bar appear.
I tried to use existing images as backgrounds. But it they all are not good for palette IMHO. So I just show palette on a dark bevel.

To close palette - use right mouse button, close button, Esc hotkey or choose any object from palette.

Here how it looks:

image

Added new hotkey attribute: allow to override common hotkey. Its needed to be able to override middle mouse button (or any other common key) in any specified section (faMapEdit/faGame/faSpectate). Middle mouse button is used for drag scroll by default, but in MapEditor - it is used for objects palette, while still can be used for drag scrolling in the game or while spectating.

This PR use #340. So please merge #340 first.

**_Update_
I tried again with background, and this time I find it very nice:

with bg

So I commited last version with background


src/KM_Controls.pas
```diff
@@ -1702,15 +1725,7 @@ procedure TKMControl.DoClick(X,Y: Integer; Shift: TShiftState; Button: TMouseBut
       fTimeOfLastClick := TimeGet;
 
     if Assigned(fOnClickShift) then
-    begin
```
reyandme, line 1706: This code is not needed anymore, as we are adding button to Shift in KM_FormMain unit, just at the start of MouseUp event processing

src/KM_Controls.pas
```diff
@@ -982,7 +989,7 @@   TKMDropCommon = class(TKMControl)
     property DropUp: Boolean read fDropUp write fDropUp;
     property ItemIndex: SmallInt read GetItemIndex write SetItemIndex;
 
-    property OnShow: TNotifyEvent read fOnShow write fOnShow;
```
reyandme, line 986: Renamed for better understanding (its actually works on show list, not on show Drop control)

src/KM_Controls.pas
```diff
@@ -202,7 +208,8 @@   TKMControl = class
     property OnClickRight: TPointEvent read fOnClickRight write fOnClickRight;
     property OnDoubleClick: TNotifyEvent read fOnDoubleClick write fOnDoubleClick;
     property OnMouseWheel: TNotifyEventMW read fOnMouseWheel write fOnMouseWheel;
-    property OnFocus: TNotifyEventFocus read fOnFocus write fOnFocus;
```
reyandme, line 205: TNotifyEventFocus is equal to existing TBooleanEvent. Removed.

src/KM_Controls.pas
```diff
@@ -202,7 +208,8 @@   TKMControl = class
     property OnClickRight: TPointEvent read fOnClickRight write fOnClickRight;
     property OnDoubleClick: TNotifyEvent read fOnDoubleClick write fOnDoubleClick;
     property OnMouseWheel: TNotifyEventMW read fOnMouseWheel write fOnMouseWheel;
-    property OnFocus: TNotifyEventFocus read fOnFocus write fOnFocus;
+    property OnFocus: TBooleanEvent read fOnFocus write fOnFocus;
+    property OnChangeVisibility: TBooleanEvent read fOnFocus write fOnChangeVisibility;
```
reyandme, line 212: New event to be able to notify control users (classes) when visibity is changed

src/KM_Controls.pas
```diff
@@ -159,9 +162,12 @@   TKMControl = class
     property AbsLeft: Integer read GetAbsLeft write SetAbsLeft;
     property AbsTop: Integer read GetAbsTop write SetAbsTop;
     property Left: Integer read GetLeft write SetLeft;
+    property Right: Integer read GetRight;
```
reyandme, line 165: added some usefull readonly properties

Kromster80:
Looks good

"Added new hotkey attribute: allow to override common hotkey. Its needed to be able to override middle mouse button (or any other common key) in any specified section (faMapEdit/faGame/faSpectate). Middle mouse button is used for drag scroll by default, but in MapEditor - it is used for objects palette, while still can be used for drag scrolling in the game or while spectating."

Am I the only one to see the inconsistency here? We establish universal behavior just to de-universalize it for some reason. Not okay with me. Maybe we can reuse spacebar instead?

reyandme:
Ok, may be its too complicated. 
May be we can split some of the keys? F.e. we can move 'scroll map' key (middle mouse) from common into game/spectate/maped. 
Other usefull hotkey for MapEd could be universal eraser, which I implemented in #330.

Kromster80:
I'm worried about standard key layout. It MUST be good, consistent, logical.
Customized layout is in players/mapmakers hands.

reyandme:
> Am I the only one to see the inconsistency here? We establish universal behavior just to de-universalize it for some reason. Not okay with me. Maybe we can reuse spacebar instead?

I agree, its too complicated. Spacebar looks good.

reyandme:
As agreed, reverted unnecessary compications on key bindings ('allow to override common hotkey' feature). 
Set spacebar as default key for objects palette.
@Kromster80 Is it ok now? 
Also, to have less diff please apply #340 first

PR #358, opened 2017-03-18
Controls: Numedit improvements and refactoring

reyandme:
1. Added new abstract control: TKMSelectableEdit. Its parent control for TKMEdit and TKMNumericEdit. Basically I just moved code, related to selecting in edit from TKMEdit to TKMSelectableEdit and reused it for TKMNumericEdit.
2. Added change NumEdit value on mouse wheel.
3. Added new ClickHold event for controls. For numeric edit its used for inc/dec value. Also we will be able to add implementaion for next task in our buglist:
'Clicking and holding on the add/subtract buttons in weapons workshops could increase the number to save repeated clicking'.

### DIFF
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -485,6 +485,7 @@ type
   end;
 
 
+  //Selectable Edit - Basic Edit class with selection available
   TKMSelectableEdit = class(TKMControl)
   private
     fFont: TKMFont;
@@ -3375,7 +3376,6 @@ procedure TKMNumericEdit.SetValue(aValue: Integer);
 begin
   fValue := EnsureRange(aValue, ValueMin, ValueMax);
   fText := IntToStr(fValue);
-  SetCursorPos(GetMaxLength);
 
   //External Value assignment should not generate OnChange event
 end;
--- a/src/KM_Controls.pas
+++ b/src/KM_Controls.pas
@@ -561,41 +561,6 @@ type
   end;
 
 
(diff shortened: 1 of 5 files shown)


reyandme:
Update state of all controls, even without any operations in TKMControls.UpdateState method seem to be not optimal and can cause small freezing while scrolling viewport.
Probably better solution could be registering controls, who need to update their state.

Kromster80:
I'm not sure ClickHold is a good solution for the problem.
It would be better to leave number-editing logic inside and present the game GUI with just the OnChange event.

What are the uses where ClickHold could be needed?

reyandme:
This event open new possibility - trigger something on mouse hold on any control. Not sure now how it could be used in the future. But for now we can use it in 2 controls: num edit (inc/dec value) and TWaresRow (or similar), where player inc/dec orders (as I mentioned, it was a note in our buglist about it, sounds reasonable). Now if we use mouse wheel orders has custom increment +-5. These increments are set in GUI class. 
@Kromster80 What are the cons of this approach ?

Kromster80:
Cons are that we exposing more methods that really needed. It is better to move in to different direction and incapsulate the internal behavior inside the control. FWIW external world does not care how NumEdit works inside, it just needs a convenient way of getting a numeric value from user.

reyandme:
Well we just create NumEdit and that's it. External world has no idea about ClickHold.
If we want we can handle it in fashion we want. 
There is no interfere with OnChange.

PR #360, opened 2017-04-17
Add the ScriptValidatorCLI util

LauraRozier:
Created to use as validator for a website that has a scripting 'playground'
Tested with Wine and that works perfectly fine.

### DIFF
--- a/KaMProjectGroup.groupproj
+++ b/KaMProjectGroup.groupproj
@@ -66,6 +66,9 @@
         <Projects Include="Utils\ScriptingParser\ScriptingParser.dproj">
             <Dependencies/>
         </Projects>
+        <Projects Include="Utils\ScriptValidatorCLI\ScriptValidatorCLI.dproj">
+            <Dependencies/>
+        </Projects>
     </ItemGroup>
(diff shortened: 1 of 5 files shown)


Kromster80:
I'm not convinced this should be a standalone util.
Why not add the CLI to existing ScriptValidator instead?

Alexandr57:
In RAD Studio there is no CLI there FMX

LauraRozier:
@Alexandr57 I'm sure you can make Console Apps in RAD, even in Tokyo Starter. :)
@Kromster80 Will look at how to implement that properly. (Already have an idea)

LauraRozier:
New repo + PR will follow shortly.

Alexandr57:
Не знаю как это работает но если через батник написать так C:\Games\KaM Remake\Campaigns\An Empire Destroyed\AED01\AED01.script
Возникает ошибка.

I do not know how it works, but if I write through the batch file as C: \ Games \ KaM Remake \ Campaigns \ An Empire Destroyed \ AED01 \ AED01.script
An error occurs.

image

PR #369, opened 2017-06-30
BugFix: demolish wrong house by using house hotkey

reyandme:
Bug description: 
1st set a hotkey to a school
2nd start a deleteing process on a coalmine but when the question is popping 'Do you really want to delete the building' then
3rd press the hotkey to the school and then the question will adressed to the school
4th now press i want to delete the building

Result is the scholl destroyed and not the coalmine.

A solution could be: If the question is up then the hotkey should overwrite the question with say 'no' and open the school or any hotkeyed building without destroying.


Kromster80:
Good fix, thanks!

PR #390, opened 2017-08-16
Replay spec clear beacons on player switch

reyandme:
This PR use fixes from #388 and #389, so its better to apply them first to make less diff here.


Kromster80:
Please resolve conflicts with previous fixes

reyandme:
Conflicts resolved

PR #391, opened 2017-08-16
GameStats(F3): added +xx for training units. Added highlight unit qty number when there is no enought workers

reyandme:
Highlight unit qty by font type - fnt_Outline, instead of fnt_Grey. 
I tried different font colors for fnt_Grey, but all of them was not so bright and not easy to find.
fnt_Outline font has small outline (suprisingly ^) so it looks much better.

Here are the examples: 

highlight


reyandme:
Added thread on [kam.net](http://www.knightsandmerchants.net/forum/viewtopic.php?f=24&t=3483)

reyandme:
PR updated.
As discussed on kam.net forum:
We count only constructed houses - for planned and under construction houses there is no need for worker, because it could be planned in advance house, or just 'saving space' from woodcutter or smth else.
There were 2 good options for color:
1. Orange:

img

2. Red:

img

I like orange - its good enought for me.

Kromster80:
Looks okay to me too. @lewinjh what do you say?

lewinjh:
The orange option looks good to me too :)

Kromster80:
+1 for orange. Red is much too aggressive

reyandme:
So if everybody agreed for orange and it is orange in my last commit can smb merge it ?

PR #392, opened 2017-08-20
Replay/Spectate - separated and sorted by teams list of hands

reyandme:
Also added new funtion to get teams from hands alliances info

Here how it looks like:

image

### DIFF
--- a/src/KM_CommonTypes.pas
+++ b/src/KM_CommonTypes.pas
@@ -8,6 +8,7 @@ uses
 type
   TKMByteArray = array of Byte;
   TKMByte2Array = array of array of Byte;
+  TKMByteSetArray = array of set of Byte;
   PKMByte2Array = ^TKMByte2Array;
   TKMWordArray = array of Word;
   PKMWordArray = ^TKMWordArray;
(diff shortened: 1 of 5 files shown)


Kromster80:
This takes too much space IMO. Why not make just a plain list?

Team1 - Player1
Team1 - Player3
Team2 - Player4
Team2 - Player5
Team3 - Player6

reyandme:
I think separating looks better, than list with 2 columns. Its easier to read - eye can separate teams visually without finding and reading team label. Its logical dividing of monotone space.

Also 4 teams is the maximum, usually its 2 teams only.
 
And if we write Team 1 - Player 1 it will also take lots of space - not in height, but in width.

Another option - do not write "Team 1" - make just plain separator, lets say 1-5px height.

img

Also we can ask on kam.net, as always, I've already create topic there.

reyandme:
I've made separators 5px height - as on screenshot above

Kromster80:
About the separators, can you make them darker and without white outline using existing methods?

reyandme:
Here it is:

img

lewinjh:
Looks good to me, it will make it easier to see what is going on in replays

PR #394, opened 2017-08-20
mk_FPS can be sent to other players only, no need to send it to ourselfs

reyandme:
This packet is sent every second, we can save bit of network traffic then


lewinjh:
Good idea. You might want to add this line after setting our local FPS:
if Assigned(fOnPingInfo) then fOnPingInfo(Self);

Otherwise it will not be updated in the GUI.

reyandme:
Yep, sure

PR #403, opened 2017-08-24
BugFix - house collection keep link to house after deleting. 

reyandme:
So even without houses you can get Houses.Count > 0 and Hand.HasAssets wil be true, you will see hand center screen/can make ally etc etc etc

### DIFF
--- a/src/hands/KM_HandsCollection.pas
+++ b/src/hands/KM_HandsCollection.pas
@@ -616,10 +616,14 @@ end;
 //MapEd procedure to remove any house under cursor
 procedure TKMHandsCollection.RemAnyHouse(Position: TKMPoint);
 var
-  I: Integer;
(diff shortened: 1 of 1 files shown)


Kromster80:
Looks like more refactoring/simplification could be done here.

reyandme:
I do not think refactoring is a good thing to do while bug fixing. Refactoring could produce new bugs and it will need much more time for testing.
Its better to do in a separate branch.

Kromster80:
Isn't there a way to use gHands.HitTestHouse(Position) or alike and delete the house properly only once, instead of keeping using the inefficient and lengthy loop?

reyandme:
I missed that HousesHitTest method - too many HitTest's everywhere.
It was not used before this PR

PR #408, opened 2017-08-25
Bugfix: lobby, select map, Map.OnChange could trigger twice

reyandme:
OnCellClick has to handle click event in all cases, not only when clicked on 1st column


reyandme:
Need to look at this problem one more time - occasionally this solution makes maps list not closing after map selecting

PR #413, opened 2017-09-06
DoTiles render performance optimisation

reyandme:
Do not bind same texture again, it will drastically change render performance

F.e. on an average Map (Cube 256x256) when full map is shown in viewport
there are only ~1250 new texture binds, when all other ~64k binds can be skipped.

Measured performance on my PC: 
for ~40k tiles DoTiles was executed in ~41ms. 
After optimisation - downto ~11ms.
from 12fps upto 20fps

for ~65k tiles in viewport
from 69ms downto 22ms
from ~7.5fps upto 12


lewinjh:
Nice work!

PR #415, opened 2017-09-06
Set atlas size depending of video adapter. But not more then 2048

reyandme:
We were using atlas size = 512, which is quite small. Its worth considering to use bigger atlases.
We can get this info from video adapter to be sure what texture size is supported.

Also may be some atlas bin packing improvement can be done:
This is what we have now, when atlas size set to 2048x2048:

img

GUI_Base and GUIMain , as well as Custom_Base atlases are not packed well. May be its not important, but anyway...

### DIFF
--- a/src/KM_BinPacking.pas
+++ b/src/KM_BinPacking.pas
@@ -29,9 +29,9 @@ type
     fImageID: Word; //Image that is using this bin (0 if unused)
     fRect: TBinRect; //Our dimensions
     fPad: Byte;
-    fNotFit: Word; //Minimum size that does not fit
+    fNotFit: Cardinal; //Minimum size that does not fit
   public
-    constructor Create(aRect: TBinRect; aPad: Byte; aImageID: Word);
+    constructor Create(aRect: TBinRect; aPad: Byte; aImageID: Word; aNotFit: Cardinal);
     destructor Destroy; override;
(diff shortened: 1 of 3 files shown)


lewinjh:
Good idea too. Have you checked how much this improves performance? Just in case it's actually slower to use larger textures (I expect this change will be faster though). I doubt any video adapters these days don't support 2048 textures.

I'm not sure why the bin packing is not working correctly for GUI_Base and GUI_Main. Looks like a bug to me. Maybe it doesn't understand your new atlas size so it puts the large sprites in a single atlases each?

reyandme:
Yes, I found that bug - there was a limit, fNotFit, set to 256x256, so bigger images always initiated new atlas.
So I can just set it to smth like AtlasSize*AtlasSize, and it will be fine. 

Also rxx packages with only 1 image (like all Campaigns) should make smaller atlases. So in case 2048*2048 there will be no too big atlases, just 1024x1024.

Also maybe we can specify atlas size for some rxx. F.e. for tiles we do not need 2048 now, 1024 is enought. 

here how it looks now:

img

reyandme:
I tried some test with lots of units and AtlasSize 4096, when all units can be placed on 1 atlas, but did not get any noticable performance improvement.
But still, if we will have atlas size = 1024x1024, then we will be able to use VBO for tiles, and that should give performance improvement

PR #420, opened 2017-09-29
BugFix: serf could bring wood/stone to worker too early

reyandme:
After some changes we order delivery earlier, than it was in r6720, so its possible, that serf will bring wood/stone too early, when worker did not finish his digging phase.
We have to ask serf to wait a bit, until worker is ready to use resource.

Bug:
https://puu.sh/xM07S/5c8df7954d.mp4

After fix:
https://puu.sh/xLZNG/83bfa6d1a0.mp4


Kromster80:
Great, thanks!

PR #422, opened 2017-10-04
Fix dynamic script compiler error caused by "array of Boolean"

LauraRozier:
This is only needed until PS is updated to where it includes the fix.

Signed-off-by: Thimo Braker <thibmorozier@gmail.com>

### DIFF
--- a/src/scripting/KM_Scripting.pas
+++ b/src/scripting/KM_Scripting.pas
@@ -228,7 +228,6 @@ begin
 
     // Common
     Sender.AddTypeS('TIntegerArray', 'array of Integer'); //Needed for PlayerGetAllUnits
-    Sender.AddTypeS('TKMBooleanArray', 'array of Boolean'); //Needed for TKMScriptUtils
     Sender.AddTypeS('TByteSet', 'set of Byte'); //Needed for Closest*MultipleTypes
 
     // Register classes and methods to the script engine.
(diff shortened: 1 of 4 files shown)


src/scripting/KM_Scripting.pas
```diff
@@ -492,12 +493,12 @@ function TKMScripting.ScriptOnUses(Sender: TPSPascalCompiler; const Name: AnsiSt
     RegisterMethodCheck(c, 'function AbsS(aValue: Single): Single');
 
     RegisterMethodCheck(c, 'function ArrayElementCount(aElement: AnsiString; aArray: array of AnsiString): Integer');
-    RegisterMethodCheck(c, 'function ArrayElementCountB(aElement: Boolean; aArray: array of Boolean): Integer');
+    RegisterMethodCheck(c, 'function ArrayElementCountB(aElement: Boolean; aArray: TKMBooleanArray): Integer');
```
Kromster80: It's not very clear if mapmakers should be using "TKMSmthArray" or "array of Smth". Let's choose one pattern and adhere to it?

Kromster80 (changes requested):

src/scripting/KM_Scripting.pas
LauraRozier: For this "temp" fix I simply changed to this.
I think changing all "array of {Type}" to TKM{Type}Array would be OK too.

I personally prefer changing to TKM{Type}Array

Kromster80: I guess "array of Smth" looks simpler and more familiar to Pascal scripters

LauraRozier: That's true.
It's also documentation-friendly. (If people search for PS they can find array of, not TKM)
What we could do is manually implement the fix for PS.

This would solve the issue as well.

Kromster80: True

PR #424, opened 2017-10-22
Terrain loading stack overflow fix

LauraRozier:
There was a stack overflow when loading bigger maps and the town tutorial.
(Could be compiler magic/specific, I used XE10.2 when it happened.)


reyandme:
@thibmo I also merged your commits into my repository, thx

PR #425, opened 2017-10-28
Fix a strange bug causing TileOverlay to stay to_none for roads

LauraRozier:
The other overlays seem to load just fine.
This issue only occurred when loading the map to play and in maped.
The preview in the menu loaded just fine.

Again, might be compiler specific.


LauraRozier:
Let's put this on-hold and perhaps use it for refactoring.

PR #428, opened 2017-11-12
Updating the UpdateList function in the RXX Editor

Alexandr57:
Made a small update for the UpdateList function.
Now the selection of the line will not vanish.


Alexandr57:
I'll do it. That I did not press something.

PR #429, opened 2017-11-12
Updating the UpdateList function in RXXEditor

Alexandr57:
Updated the UpdateList function. Now the exclusion will not merge.


Kromster80:
Variable name is non-standard. "a" prefix is for method arguments.
Range checking is too complcated. Should have been simple "ItemIndex := EnsureRange(..)"

Issue #440, opened 2018-09-19
AVI playback


Kromster80:
The game requires AVI playback for intro and inter-mission videos.

Requirements for a solution:
 - support for overlay text/subtitles
 - support different voice-overs
 - de-interlace on the fly (or we could re-code de-interlaced video)
 - compile on Delphi7-XE5 and Lazarus
 - ensure that codec used is available on all systems (for original codecs - looks like it is)
 - any library used needs to be licensed for free use
 - video and subtitles should look okay with dynamic resize (1024*720 to 4k) and different windowed frame proportions

--

Considerations:
 - prefer to use OpenGL renderer with POT textures
 - subtitles will probably need a new font (or some system font), to be large enough (readable)
 - make calls to play mission videos a part of campaign data

PR #442, opened 2018-09-20
Added K&M AVI Player

stucuk:
Added K&M AVI Player i made to Utils folder.


Kromster80:
Issues that need to be fixed:
 - exclude DCU, DDP, RES files
 - DSK file seems to be not needed either?
 - change project structure to Utils\project_files and Utils\3rdparty\for_libraries. So that there's no extra folders in path

LauraRozier:
ref #440

Added comment to keep track of the issue related to this PR

PR #451, opened 2019-08-05
refine/add Chinese translations.

XenoAmess:
refine Chinese translations in menu (implement some and fix some)

add translation of the 5 campaigns, and two tutorials.

translate 4 maps.(too many maps so I just cannot translate them all...)

I am sure I can read them in my editors(vscode / idea).
But I cannot test if it can run correctly (due to transcode or something) because I don't know how to compile and test it.

I'm not a translation specialist but I'm Chinese, so I can guarantee the translations are at least able to read (though not very beautiful lol).


XenoAmess:
@reyandme The Chinese translation for compaigns I promised several months ago. I finally get some time and finished this lol.

PR #456, opened 2021-05-02
update using RAD Studio 10.4

AndreiCherniaev:


Kromster80:
WDC is defined in the KaM_Remake.inc, no need to fix it "ad hoc" =)

